Issue #8, opened 2022-05-12
Static content changes not reflected in preview channel


ErikWittern:
I am trying to use a preview channel when deploying a Next.js application. Helpfully, the CLI warns me that preview channels will not work with the generated Cloud Function (yet? Support for this would be fantastic!):

```
Cannot preview changes to the backend, you will only see changes to the static content on this channel.
```

Despite the warning, and me continuing the deploy anyways, the static content is not updated in the created preview channel. Instead, the same content as deployed to the main channel is served under the preview channel's URL. This is despite the page in question (`versions`) being marked as `static` by the Next.js build log:

```
└ ○ /versions                              329 B          76.1 kB

...

○  (Static)  automatically rendered as static HTML (uses no initial props)
```

jamesdaniels:
Interesting… we’ll investigate.

ErikWittern:
@jamesdaniels Please let me know if I can be of further help, i.e., provide more information about my setup or similar.

jamesdaniels:
If I were to guess, it’s that your pre-rendered pages are transformed into paths and .html files in a way that aren’t getting picked up by hosting, so all the content is falling back to the Cloud Function (which doesn’t support previews ATM), rather than served by the CDN. Does that pass the sniff test @ErikWittern? 

If you don’t mind zipping up your generated .firebase directory and sending it my way, jamesdaniels@google.com. I’ll take a look when I return from my leave in June.

ErikWittern:
@jamesdaniels Your theory sounds very plausible. The HTML files contained `.firebase/{projectId}/hosting` after deploying the preview channel contain the updated content, but the content served from the preview channel still shows the content last deployed to the main channel.

I sent you the zipped `.firebase` channel for further investigation.

jamesdaniels:
The root cause here was the cleanUrls setting. cleanUrls is now default true for web frameworks in the[ latest version of the preview integration](https://firebase.google.com/docs/hosting/nextjs).

PR #16, opened 2022-05-31
Add common configurations to README.md

aabccd021:
In my next.js environment, after I created a page in `pages/foo/bar.tsx`, the page was only accessible through url `/foo/bar.html`.
I was able to access it through `/foo/bar` after adding `"cleanUrls": true` to my `firebase.json` hosting config.

my `next.config.js`:
```js
/** @type {import('next').NextConfig} */
const nextConfig = {
  reactStrictMode: true,
}

module.exports = nextConfig
```

my `package.json`:
```json
{
 "engines": {
    "node": "16"
  },
  "scripts": {
    "dev": "firebase emulators:start",
    "postinstall": "firebase --open-sesame frameworkawareness"
  },
  "dependencies": {
    "firebase": "^9.8.1",
    "firebase-admin": "^10.0.2",
    "firebase-functions": "^3.21.2",
    "firebase-tools": "^10.9.2",
    "next": "12.1.6"
  },
  "devDependencies": {
    "@types/react": "18.0.9",
    "@types/react-dom": "18.0.4",
    "firebase-functions-test": "^0.2.0",
    "typescript": "4.6.4"
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,22 @@ When you're ready to share your changes with the world, deploy your Next.js app 
 1. Run `firebase deploy` from the terminal.
 2. Check your website on: `SITE_ID.web.app` or `PROJECT_ID.web.app` (or your custom domain, if you did setup one)
 
+### Common Configurations
+
+#### Add `cleanUrls` option
+
+By default, a page created on `/pages/foo/bar.jsx` is only accessible through the url `/foo/bar.html`.
+To make the page accessible on `/foo/bar` instead, add `cleanUrls` option to your hosting config in firebase.json.
+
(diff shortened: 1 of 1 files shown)


jamesdaniels:
Ah, maybe this is the root cause of #8. Perhaps we should automatically add cleanUrls?

aabccd021:
> Ah, maybe this is the root cause of #8. Perhaps we should automatically add cleanUrls?

Sounds good to me

jamesdaniels:
In the meantime I'll merge. Thanks for catching that.

PR #30, opened 2022-08-26
Vite build support

davideast:


package.json
```diff
@@ -99,6 +100,7 @@
     "rimraf": "^3.0.2",
     "ts-node": "^10.9.1",
     "typescript": "^4.7.4",
+    "vite": "^2.9.15",
```
davideast, line 103: Is this needed since we are relying on the developer's version?

jamesdaniels: we want it as a dev dep for the types, yes

jamesdaniels:
Closing in favor of https://github.com/firebase/firebase-tools/pull/4868

PR #47, opened 2022-10-17
Drop Next minimal mode, fixes image optimization

jamesdaniels:
Next.js' minimal mode is unstable, I was using it to opt-out of ISR and touching the filesystem in Cloud Functions but it seems it also disables Image Optimzation—so drop it. There is no benefit to us ATM as it seems to be for edge compute rather than serverless.

* Drop minimal mode
* Bump version
* Fix firebase-tools devDep for test suite

This is purely contained within the firebase-frameworks NPM dep, no firebase-tools patch is required to take advantage.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -97,7 +97,7 @@ jobs:
         run: mv firebase-frameworks-${{ github.run_id }} dist
       - name: Test
         run: |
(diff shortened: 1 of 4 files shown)


jamesdaniels:
FYI nuxt failures in the test suite can be ignored ATM

TheIronDev (approved):

PR #56, opened 2022-12-20
Add basic SvelteKit support

austincrim:
Tested with SvelteKit demo app, which includes static and server-side rendered pages, API routes, and form actions.


src/sveltekit/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const body = await rendered.text()
+  return res
+    .writeHead(rendered.status, Object.fromEntries(rendered.headers))
+    .end(body)
+}
+
+// https://github.com/jthegedus/svelte-adapter-firebase/blob/main/src/files/firebase-to-svelte-kit.js
+function toSvelteKitRequest(request: Request) {
+  // Firebase sometimes omits the protocol used. Default to http.
```
jamesdaniels, line 35: gen2 functions should handle x-forwarded correctly, so these links shouldn't be needed any longer

src/sveltekit/index.ts
```diff
@@ -0,0 +1,57 @@
...
+import type { IncomingHttpHeaders } from 'http'
+
+const kitPromise = import(`${pathToFileURL(process.cwd())}/index.js`)
+const manifestPromise = import(`${pathToFileURL(process.cwd())}/manifest.js`)
+
+// Request, Response, fetch, etc.
+installPolyfills()
+
+export const handle = async (req: Request, res: Response) => {
+  const [{ Server }, { manifest }] = await Promise.all([
```
jamesdaniels, line 14: can you pull lines 14-20 outside of the handler, so it used across requests

jamesdaniels (changes requested):

jamesdaniels (changes requested):

src/sveltekit/index.ts
austincrim: From what I can tell, this seems to be on the only way to get the actual full `host` to pass as the base in the `URL` constructor that follows. I couldn't find a way to piece  together the `host` without using these headers.

austincrim:
The [deployed demo app](https://fb-tools-dev.web.app/sverdle) is not totally functional yet. The Wordle clone doesn't work, I think due to some issues with cookies. It works fine locally, just seeing issues when deployed. I'm investigating further.

PR #78, opened 2023-05-08
Check token expiry instead of auth_time

ProfHercules:
I might be missing something, but checking `auth_time` always seems to return the first time the user authenticated, which means the token is always expired after 5 mins.


jamesdaniels:
I think IAT is what I really intended, will address in the next release. Thanks for taking the initiative here.

PR #84, opened 2023-06-23
Fix `usingFirebaseJsSdk`

leoortizz:
Apart from `process.env.__FIREBASE_DEFAULTS__`, check whether the Firebase JS SDK can be imported to assume that the app is using it. In some cases the env. variable was set even if the app was not using the Firebase JS SDK.

Fixes https://github.com/firebase/firebase-tools/issues/5996

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,0 +1,11 @@
+export async function isUsingFirebaseJsSdk() {
+  if(!process.env.__FIREBASE_DEFAULTS__) return false;
+
+  try {
+    await import('firebase/app');
+
+    return true
+  } catch (e) {
+    return false;
+  }
+}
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -0,0 +1,11 @@
...
+  if(!process.env.__FIREBASE_DEFAULTS__) return false;
+
+  try {
+    await import('firebase/app');
+
+    return true
+  } catch (e) {
+    return false;
+  }
+}
```
alexastrum, line 11: Nit: line break at end of files.

alexastrum (approved):

src/utils.ts
leoortizz: I see, there's no prettier config in this repo though, we should add it, probably use the same config as the CLI

PR #85, opened 2023-06-24
Fix formatting

alexastrum:


leoortizz:
@alexastrum what was used for formatting as there's no Prettier config in this repo? I'd like to follow it for future commits

alexastrum:
> @alexastrum what was used for formatting as there's no Prettier config in this repo? I'd like to follow it for future commits

I added a prettier config similar to the one from `firebase-tools` main repo.

PR #91, opened 2023-07-13
Fix Nuxt 3 tests

leoortizz:
Rename Nuxt 3 from `nuxt3` to `nuxt`, and Nuxt 2 from `nuxt` to `nuxt2`. 

The Nuxt 3 test was having errors in the `postinstall` script that runs `nuxt prepare` due to the `nuxt` dependency in the parent directory. Found that this could be the issue from this comment: https://github.com/nuxt/nuxt/issues/19866#issuecomment-1486414872


alexastrum:
@leoortizz, pls run `npm i` and push latest `package-lock.json`

> npm ERR! `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.

PR #92, opened 2023-08-03
allow firebase sdk v10 in peer deps

austincrim:
fixes https://github.com/firebase/firebase-tools/issues/6109

**Question:** After we cut a new release, I thought we might need to bump the [`FIREBASE_FRAMEWORKS_VERSION` constant](https://github.com/firebase/firebase-tools/blob/v12.4.3/src/frameworks/constants.ts#L28) in `firebase-tools`, but since it has the `^`, it should download the new `firebase-frameworks` version right?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
         "node": "^16.0.0 || ^18.0.0"
       },
       "peerDependencies": {
-        "firebase": "^9.9.0",
+        "firebase": "^9.0.0 || ^10.0.0",
         "firebase-admin": "^11.0.1",
         "sharp": "^0.32.1"
       },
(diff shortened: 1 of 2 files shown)


taeold (approved):

jhuleatt (approved):

austincrim:
@jhuleatt I'm not sure why only the custom tests are failing, my changes should not impact them in any way.

I did notice the same tests are failing on [my colleague's PR](https://github.com/FirebaseExtended/firebase-framework-tools/pull/91), maybe it's related to making a PR off of a fork?

alexastrum:
`custom` test seems to be failing on `main`, so it's for a different reason.

MeTaNoV:
It is still an issue with firebase-tools v12.8.0 ...

PR #95, opened 2023-08-09
Enable next.js test action

alexastrum:


alexastrum:
This fixes and enables `next` tests and seems to pass `custom` tests as well, even though they were failing on local branches. So I want this pulled into existing feature branches ASAP to make sure we don't introduce new test failures.

PR #103, opened 2023-08-24
Allow Firebase JS SDK >= 9.9.0

leoortizz:
#92 allowed Firebase `^10.0.0`, but `10.1.0` errors out as it doesn't match the expected version.

I changed it to `>=9.9.0`, as `9.9.0` was the version defined before #92


jhuleatt:
`^10.0.0` should include `10.1.0` per the [npm semver calculator](https://semver.npmjs.com/). Is something else going on that is causing this?

jhuleatt:
Oh, I think the issue is just that we never released it

leoortizz:
> Oh, I think the issue is just that we never released it

oh, might be it! I have not checked which version was being used when I had the error.

alexastrum (approved):

leoortizz:
@jhuleatt @alexastrum closing this PR as the real issue is that #92 was merged but was never released to NPM, so people still having the peer dependency issue with `firebase` >= 10.

MeTaNoV:
it is not fixed yet...

PR #119, opened 2023-11-20
Update adaptor build script

sjjj986:
Nextjs adaptor build script

### DIFF
--- a/packages/@apphosting/adapter-nextjs/src/bin/build.ts
+++ b/packages/@apphosting/adapter-nextjs/src/bin/build.ts
@@ -37,7 +37,7 @@ function build(cwd: string) {
     spawnSync("npm run", ["build"], {cwd, shell: true, stdio: "inherit"}); 
   }
 
-// move public directory to the public directory in apphosting output directory
+// move public directory to apphosting output public directory
 const movePublicDirectory = async () => {
     const publicDirectoryExists = await exists(appHostingPublicDirectory);
     if (!publicDirectoryExists) return;
--- a/packages/@apphosting/adapter-nextjs/src/bin/build.ts
+++ b/packages/@apphosting/adapter-nextjs/src/bin/build.ts
@@ -6,6 +6,7 @@ import { join } from "path";
 import fsExtra from "fs-extra";
 import { stringify as yamlStringify } from "yaml";
 
+// unable to use shorthand imports on fsExtra since fsExtra is CJS
 const { move, exists, writeFile, mkdirp } = fsExtra;
 const cwd = process.cwd();
 
@@ -16,30 +17,31 @@ process.env.NEXT_TELEMETRY_DISABLED = "1";
 
 build(cwd);
 
-const {distDir, basePath} = await loadConfig(cwd);
+const {distDir} = await loadConfig(cwd);
 const manifest = await readRoutesManifest(join(cwd, distDir));
 
-const destination = join(cwd, ".apphosting");
-const staticDestination = join(destination,"_next", "static");
-const publicDestination = join(destination, "public");
-const outputBundleDestination = join(destination, "bundle.yaml");
+const appHostingOutputDirectory = join(cwd, ".apphosting");
+const appHostingStaticDirectory = join(appHostingOutputDirectory,"_next", "static");
+const appHostingPublicDirectory = join(appHostingOutputDirectory, "public");
+const outputBundlePath = join(appHostingOutputDirectory, "bundle.yaml");
+const serverFilePath = join(appHostingOutputDirectory, "server.js");
 
-const standaloneDir = join (cwd, distDir, "standalone");
-const staticDir = join (cwd, distDir, "static");
-const publicDir = join(cwd, "public");
+const standaloneDirectory = join(cwd, distDir, "standalone");
+const staticDirectory = join(cwd, distDir, "static");
+const publicDirectory = join(cwd, "public");
 
-await mkdirp(staticDestination);
+await mkdirp(appHostingStaticDirectory);
 
 // Run build command
 function build(cwd: string) {
-    spawnSync("npm run", ["build"], {cwd, stdio: "inherit"}); 
+    spawnSync("npm run", ["build"], {cwd, shell: true, stdio: "inherit"}); 
   }
 
-// move public directory to both public and CDN destination
-const movePublicDir = async () => {
-    const publicDirExists = await exists(publicDir);
-    if (!publicDirExists) return;
-    await move(publicDir, publicDestination);
+// move public directory to the public directory in apphosting output directory
+const movePublicDirectory = async () => {
+    const publicDirectoryExists = await exists(appHostingPublicDirectory);
+    if (!publicDirectoryExists) return;
+    await move(publicDirectory, appHostingPublicDirectory, { overwrite: true });
 };
   
 // generate bundle.yaml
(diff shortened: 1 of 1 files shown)


packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -2,12 +2,58 @@
 import { spawnSync } from "child_process";
 import { loadConfig, readRoutesManifest } from "../utils.js";
 
+import { join } from "path";
+import fsExtra from "fs-extra";
+import { stringify as yamlStringify } from "yaml";
+
+const { move, exists, writeFile, mkdirp } = fsExtra;
```
jamesdaniels, line 9: nit: I'd set a comment about fsExtra being CJS and not being able to use shorthand imports.

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -2,12 +2,58 @@
...
+process.env.NEXT_PRIVATE_STANDALONE = "true"; 
+// Opt-out sending telemetry to Vercel
+process.env.NEXT_TELEMETRY_DISABLED = "1"; 
+
+build(cwd);
+
+const {distDir, basePath} = await loadConfig(cwd);
+const manifest = await readRoutesManifest(join(cwd, distDir));
+
+const destination = join(cwd, ".apphosting");
```
jamesdaniels, line 22: How about naming appHostingOutputDirectory or something more descriptive, here and for the other destination paths?

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -2,12 +2,58 @@
...
+    if (!publicDirExists) return;
+    await move(publicDir, publicDestination);
+};
+  
+// generate bundle.yaml
+const generateBundleYaml = async () => {
+    const headers = manifest.headers.map(it => ({...it, regex: undefined}));
+    const redirects = manifest.redirects.filter(it => !it.internal).map(it => ({...it, regex: undefined}));
+    const beforeFileRewrites = Array.isArray(manifest.rewrites) ? manifest.rewrites : manifest.rewrites?.beforeFiles || [];
+    const rewrites = beforeFileRewrites.map(it => ({...it, regec: undefined}));
```
jamesdaniels, line 50: `regex: undefined`?

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -2,12 +2,58 @@
...
+
+const standaloneDir = join (cwd, distDir, "standalone");
+const staticDir = join (cwd, distDir, "static");
+const publicDir = join(cwd, "public");
+
+await mkdirp(staticDestination);
+
+// Run build command
+function build(cwd: string) {
+    spawnSync("npm run", ["build"], {cwd, stdio: "inherit"});
```
jamesdaniels, line 35: Move run into the second arg, `"npm", ["run", "build"]...`

Also my spidey sense tells me this might need `shell: true` but I'm not sure if that's my being overly defensive, worth investigating e.g, `spawnSync("npm", ["run", "build"], {cwd, shell: true, stdio: "inherit"});`

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -2,12 +2,58 @@
...
+await mkdirp(staticDestination);
+
+// Run build command
+function build(cwd: string) {
+    spawnSync("npm run", ["build"], {cwd, stdio: "inherit"}); 
+  }
 
-const config = await loadConfig(cwd);
-const routeManifest = await readRoutesManifest(config.distDir);
+// move public directory to both public and CDN destination
```
Yuangwang, line 38: This code below just moves the public directory to the public directory under .apphosting right? I don't think CDN destination is involved.

Also for now don't worry about moving anything to a CDN directory I think this logic is good already

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -2,12 +2,58 @@
...
+    await move(publicDir, publicDestination);
+};
+  
+// generate bundle.yaml
+const generateBundleYaml = async () => {
+    const headers = manifest.headers.map(it => ({...it, regex: undefined}));
+    const redirects = manifest.redirects.filter(it => !it.internal).map(it => ({...it, regex: undefined}));
+    const beforeFileRewrites = Array.isArray(manifest.rewrites) ? manifest.rewrites : manifest.rewrites?.beforeFiles || [];
+    const rewrites = beforeFileRewrites.map(it => ({...it, regec: undefined}));
+    await writeFile(outputBundleDestination, yamlStringify({headers, redirects, rewrites}))
```
Yuangwang, line 51: Along with these fields could you also pass along the following in bundle.yaml?

1.  runCommand - type: String, the command that will run the server
2. neededDirs - type: String array, the minimum directories we need to run the server
3. staticAssets - type: String array, the directory locations of the static assets

In this specific case, I'm expecting the bundle.yaml to look like the following:

bundle.yaml:
```
...
runCommand: node ./.aphosting/server.js
neededDirs:
  - .apphosting
staticAssets:
  - /public
  - /.next/static
...
```
This will be needed in the surrounding buildpack steps

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -2,12 +2,58 @@
...
+// Opt-out sending telemetry to Vercel
+process.env.NEXT_TELEMETRY_DISABLED = "1"; 
+
+build(cwd);
+
+const {distDir, basePath} = await loadConfig(cwd);
+const manifest = await readRoutesManifest(join(cwd, distDir));
+
+const destination = join(cwd, ".apphosting");
+const staticDestination = join(destination,"_next", "static");
```
Yuangwang, line 23: Why _next instead of .next?

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -2,12 +2,58 @@
...
+    const redirects = manifest.redirects.filter(it => !it.internal).map(it => ({...it, regex: undefined}));
+    const beforeFileRewrites = Array.isArray(manifest.rewrites) ? manifest.rewrites : manifest.rewrites?.beforeFiles || [];
+    const rewrites = beforeFileRewrites.map(it => ({...it, regec: undefined}));
+    await writeFile(outputBundleDestination, yamlStringify({headers, redirects, rewrites}))
+}
 
-// TODO do all the things
-console.log({ config, routeManifest });
+await Promise.all([
+    move(standaloneDir, destination),
```
Yuangwang, line 55: Move might have some issues here since the destination is already created, you probably want to add { overwrite: true } as an option. so something like

move(standaloneDir, destination, { overwrite: true });

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -2,12 +2,58 @@
...
+    const beforeFileRewrites = Array.isArray(manifest.rewrites) ? manifest.rewrites : manifest.rewrites?.beforeFiles || [];
+    const rewrites = beforeFileRewrites.map(it => ({...it, regec: undefined}));
+    await writeFile(outputBundleDestination, yamlStringify({headers, redirects, rewrites}))
+}
 
-// TODO do all the things
-console.log({ config, routeManifest });
+await Promise.all([
+    move(standaloneDir, destination), 
+    move(staticDir, staticDestination),
```
Yuangwang, line 56: same as above

packages/@apphosting/adapter-nextjs/src/bin/build.ts
sjjj986: after some research, it seems like having shell: true is indeed very safe, there could possibly be empty output without that due to syntax problems

packages/@apphosting/adapter-nextjs/src/bin/build.ts
sjjj986: good catch! forgot to remove that when I remove the CDN directory

packages/@apphosting/adapter-nextjs/src/bin/build.ts
sjjj986: I asked James this question, seems like it's just the way nextjs output path is named, we have to reverse engineer this part and use "_next" instead of ".next"

packages/@apphosting/adapter-nextjs/src/bin/build.ts
sjjj986: added

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -2,12 +2,69 @@
...
+
+const standaloneDirectory = join(cwd, distDir, "standalone");
+const staticDirectory = join(cwd, distDir, "static");
+const publicDirectory = join(cwd, "public");
+
+await mkdirp(appHostingStaticDirectory);
+
+// Run build command
+function build(cwd: string) {
+    spawnSync("npm run", ["build"], {cwd, shell: true, stdio: "inherit"});
```
Yuangwang, line 37: I think the syntax here might be wrong I think probably should be 
spawnSync("npm", ["run", "build"], {cwd, shell: true, stdio: "inherit"});
 instead

Yuangwang (approved):

PR #121, opened 2023-11-29
Apphosting dir fixes

jamesdaniels:
* For standalone and no CDN static directory should be placed in .next ([see docs here](https://nextjs.org/docs/pages/api-reference/next-config-js/output#automatically-copying-traced-files))
* Public dir existence check was looking at the wrong spot
* Use relative directories in the bundle.yaml
* Fix the race conditions in the folder moving which were wiping files out

### DIFF
--- a/packages/@apphosting/adapter-nextjs/src/bin/build.ts
+++ b/packages/@apphosting/adapter-nextjs/src/bin/build.ts
@@ -17,20 +17,18 @@ process.env.NEXT_TELEMETRY_DISABLED = "1";
 
 build(cwd);
 
-const {distDir} = await loadConfig(cwd);
+const { distDir } = await loadConfig(cwd);
 const manifest = await readRoutesManifest(join(cwd, distDir));
 
 const appHostingOutputDirectory = join(cwd, ".apphosting");
+const appHostingStaticDirectory = join(appHostingOutputDirectory, ".next", "static");
 const appHostingPublicDirectory = join(appHostingOutputDirectory, "public");
-const nextStaticAssetsDestination = join(appHostingPublicDirectory, manifest.basePath, "_next", "static");
 const outputBundlePath = join(appHostingOutputDirectory, "bundle.yaml");
 const serverFilePath = join(appHostingOutputDirectory, "server.js");
 
 const standaloneDirectory = join(cwd, distDir, "standalone");
 const staticDirectory = join(cwd, distDir, "static");
 const publicDirectory = join(cwd, "public");
-
-await mkdirp(nextStaticAssetsDestination);
 
 // Run build command
 function build(cwd: string) {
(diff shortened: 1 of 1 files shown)


packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -50,21 +48,22 @@ const generateBundleYaml = async () => {
...
     }));
 }
 
 // move the standalone directory, the static directory and the public directory to apphosting output directory
 // as well as generating bundle.yaml
+await move(standaloneDirectory, appHostingOutputDirectory, { overwrite: true });
 await Promise.all([
-    move(standaloneDirectory, appHostingOutputDirectory, { overwrite: true }), 
-    move(staticDirectory, appHostingStaticDirectory, { overwrite: true }), 
+    await move(staticDirectory, appHostingStaticDirectory, { overwrite: true }),
```
jamesdaniels, line 66: opps need to kill the await here

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -50,21 +48,22 @@ const generateBundleYaml = async () => {
...
     const rewrites = beforeFileRewrites.map(it => ({...it, regex: undefined}));
+    const outputBundleDirectory = dirname(outputBundlePath);
     await writeFile(outputBundlePath, yamlStringify({
         headers, 
         redirects, 
         rewrites,
-        runCommand: `node ${serverFilePath}`,
-        neededDirs: [appHostingOutputDirectory],
-        staticAssets: [appHostingStaticDirectory, appHostingPublicDirectory],
+        runCommand: `node ${normalize(relative(outputBundleDirectory, serverFilePath))}`,
```
sjjj986, line 56: I checked with Wei, and his buildpack step want to use paths relative to the base directory, so suggesting changing `outputBundleDirectory` to `cwd`

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -50,21 +48,22 @@ const generateBundleYaml = async () => {
...
+        neededDirs: [normalize(relative(outputBundleDirectory, appHostingOutputDirectory))],
+        staticAssets: [normalize(relative(outputBundleDirectory, appHostingPublicDirectory))],
```
sjjj986, line 58: i think this is missing the static directory

packages/@apphosting/adapter-nextjs/src/bin/build.ts
jamesdaniels: can do. including run command? or just the directories?

packages/@apphosting/adapter-nextjs/src/bin/build.ts
jamesdaniels: static is now part of .next directory per NextJS standalone specifications, we can further optimize when we have CDN. SG?

packages/@apphosting/adapter-nextjs/src/bin/build.ts
sjjj986: including the run command

packages/@apphosting/adapter-nextjs/src/bin/build.ts
sjjj986: SG 👍

sjjj986 (approved):

PR #122, opened 2023-11-30
Possible fix for readable being spent in functions

jamesdaniels:


leoortizz (approved):
amazing! This fixes requests with body, server actions, and probably more. Thanks @jamesdaniels!

jamesdaniels:
@mfellner thanks for tracking down the problem and the inspiration, I think we have a solution here.

PR #125, opened 2023-12-06
`localhost` as Next.js server hostname

leoortizz:
fixes 500 error `"url" parameter is valid but upstream response is invalid` in images optimized by Next.js

TODO

- [ ] more testing with different Next.js versions


jamesdaniels (approved):
LGTM. Make sure to manually test against older NextJS versions.

ian-sayles:
Not sure if this as caused me my current issue, but when I now deploy my site to a channel, it does not create a cloud function like it used to do for the images to work, the deploy is a lot faster but only certain images are being shown in the app, was working the other day, updated firebase tools this afternoon.

PR #134, opened 2024-01-09
add build:watch script

leoortizz:


package.json
```diff
@@ -3,6 +3,7 @@
   "private": true,
   "scripts": {
     "build": "./scripts/build.js",
+    "build:watch": "npm run build && tsc --watch",
```
jamesdaniels, line 6: did you try `lerna watch`

leoortizz: good call, thanks. forgot about lerna 🙂

jamesdaniels (approved):

PR #139, opened 2024-01-29
angular build adatper draft

sjjj986:
angular build adapter script 

output directory structure & bundle.yaml file after running adapter:
Screenshot 2024-01-29 at 5 00 46 PM

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -13,6 +13,7 @@ module.exports = {
   ],
   rules: {
     "jsdoc/newline-after-description": "off",
+    "@typescript-eslint/no-explicit-any": "off", // TODO(sijin): remove later
     "jsdoc/require-jsdoc": "off",
     "no-restricted-globals": ["error", "name", "length"],
     "prefer-arrow-callback": "error",
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "packages/**/*"
   ],
   "devDependencies": {
+    "@angular-devkit/build-angular": "^17.1.2",
     "@types/mocha": "*",
     "@typescript-eslint/eslint-plugin": "^6.10.0",
     "eslint": "^8.53.0",
@@ -25,6 +26,7 @@
     "mocha": "^10.2.0",
     "prettier": "^3.0.3",
     "ts-mocha": "^10.0.0",
-    "typescript": "^5.2.0"
+    "typescript": "^5.2.0",
+    "zod": "^3.22.4"
   }
 }
--- a/packages/@apphosting/adapter-angular/src/bin/build.ts
+++ b/packages/@apphosting/adapter-angular/src/bin/build.ts
@@ -12,20 +12,18 @@ interface OutputBundleOptions {
   browserDirectory: string;
 }
 
-const cwd = process.cwd()
+const cwd = process.cwd();
 
 // read Application Builder config
 const config: ApplicationBuilderOptions = await loadConfig(cwd);
 const outputPath = config.outputPath;
 // normalized output path structure
-const normalizedOutputPath = { 
-  browser: 'browser',
-  server: 'server',
-  media: 'media',
-  ...(typeof outputPath === 'string' ? undefined : outputPath),
-  base: normalize(
-    resolve(typeof outputPath === 'string' ? outputPath : outputPath.base),
-  ),
+const normalizedOutputPath = {
+  browser: "browser",
+  server: "server",
+  media: "media",
+  ...(typeof outputPath === "string" ? undefined : outputPath),
+  base: normalize(resolve(typeof outputPath === "string" ? outputPath : outputPath.base)),
 };
 
 // populate file/directory paths we need inside app hosting output directory
@@ -54,7 +52,7 @@ async function generateOutputDirectory(
   const baseDirectory = resolve(normalizedOutputPath.base);
 
   await Promise.all([
-    move(baseDirectory, outputBundleOptions.outputBaseDirectory, { overwrite: true }), 
+    move(baseDirectory, outputBundleOptions.outputBaseDirectory, { overwrite: true }),
     generateBundleYaml(outputBundleOptions, cwd),
   ]);
   return;
@@ -63,7 +61,7 @@ async function generateOutputDirectory(
 // generate bundle.yaml
 async function generateBundleYaml(
   outputBundleOptions: OutputBundleOptions,
-  cwd : string,
+  cwd: string,
 ): Promise<void> {
   await writeFile(
     outputBundleOptions.bundleYamlPath,
--- a/packages/@apphosting/adapter-angular/src/bin/build.ts
+++ b/packages/@apphosting/adapter-angular/src/bin/build.ts
@@ -1,16 +1,13 @@
 #! /usr/bin/env node
 import {
-  loadConfig,
+  checkBuildConditions,
   build,
-  populateOutputBundleOptions,
   generateOutputDirectory,
 } from "../utils.js";
 
 const cwd = process.cwd();
 
-// read Application Builder config
-const config = await loadConfig(cwd);
+await checkBuildConditions(cwd);
 
-const outputBundleOptions = populateOutputBundleOptions(config);
-await build().catch(() => process.exit(1));
+const outputBundleOptions = await build().catch(() => process.exit(1));
 await generateOutputDirectory(cwd, outputBundleOptions);
--- a/packages/@apphosting/adapter-angular/src/bin/build.ts
+++ b/packages/@apphosting/adapter-angular/src/bin/build.ts
@@ -39,9 +39,14 @@ function populateOutputBundleOptions(): OutputBundleOptions {
 }
 
 // Run build command
-async function build(cwd: string): Promise<void> {
-  await spawn("npm", ["run", "build"], { cwd, shell: true, stdio: "inherit" });
-}
+const build = (cwd = process.cwd()) =>
+  new Promise<void>((resolve, reject) => {
+    const process = spawn("npm", ["run", "build"], { cwd, shell: true, stdio: "inherit" });
+    process.on("exit", (code) => {
+      if (code === 0) return resolve();
+      reject();
+    });
+  });
 
 // move the base output directory, which contains the server and browser bundle directory, and prerendered routes
 // as well as generating bundle.yaml
@@ -78,5 +83,5 @@ async function generateBundleYaml(
 }
 
 const outputBundleOptions = populateOutputBundleOptions();
-await build(cwd);
+await build().catch(() => process.exit(1));
 await generateOutputDirectory(cwd, outputBundleOptions);
--- a/packages/@apphosting/adapter-angular/src/bin/build.ts
+++ b/packages/@apphosting/adapter-angular/src/bin/build.ts
@@ -1,7 +1,7 @@
 #! /usr/bin/env node
 import { spawn } from "child_process";
 import { loadConfig, ApplicationBuilderOptions, move, writeFile } from "../utils.js";
-import { join, resolve, normalize, relative } from "path";
+import { resolve, normalize, relative } from "path";
 import { stringify as yamlStringify } from "yaml";
 
 interface OutputBundleOptions {
@@ -36,13 +36,13 @@ function populateOutputBundleOptions(): OutputBundleOptions {
     outputDirectory: outputBundleDir,
     outputBaseDirectory: resolve(outputBundleDir, "dist"),
     serverFilePath: resolve(outputBundleDir, "dist", normalizedOutputPath.server, "server.mjs"),
-    browserDirectory: resolve(outputBundleDir, "dist", normalizedOutputPath.browser)
+    browserDirectory: resolve(outputBundleDir, "dist", normalizedOutputPath.browser),
   };
 }
 
 // Run build command
 async function build(cwd: string): Promise<void> {
-  spawn("npm", ["run", "build"], { cwd, shell: true, stdio: "inherit" });
+  await spawn("npm", ["run", "build"], { cwd, shell: true, stdio: "inherit" });
 }
 
 // move the base output directory, which contains the server and browser bundle directory, and prerendered routes
--- a/packages/@apphosting/adapter-angular/src/bin/build.ts
+++ b/packages/@apphosting/adapter-angular/src/bin/build.ts
@@ -1,87 +1,16 @@
 #! /usr/bin/env node
-import { spawn } from "child_process";
-import { loadConfig, ApplicationBuilderOptions, move, writeFile } from "../utils.js";
-import { resolve, normalize, relative } from "path";
-import { stringify as yamlStringify } from "yaml";
-
-interface OutputBundleOptions {
-  bundleYamlPath: string;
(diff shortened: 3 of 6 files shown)


packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -1,20 +1,87 @@
...
+}
+
+const cwd = process.cwd();
+
+// read Application Builder config
+const config: ApplicationBuilderOptions = await loadConfig(cwd);
+const outputPath = config.outputPath;
+// normalized output path structure
+const normalizedOutputPath = {
+  browser: "browser",
```
jamesdaniels, line 22: When are we expecting that developers can override these? Should we build that in now?

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -1,20 +1,87 @@
 #! /usr/bin/env node
 import { spawn } from "child_process";
-import { loadConfig } from "../utils.js";
+import { loadConfig, ApplicationBuilderOptions, move, writeFile } from "../utils.js";
+import { resolve, normalize, relative } from "path";
+import { stringify as yamlStringify } from "yaml";
 
+interface OutputBundleOptions {
```
jamesdaniels, line 7: Can we make an interfaces.ts and export this there?

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -1,20 +1,87 @@
...
+  outputBundleOptions: OutputBundleOptions,
+): Promise<void> {
+  const baseDirectory = resolve(normalizedOutputPath.base);
 
-await build().catch(() => process.exit(1));
+  await Promise.all([
+    move(baseDirectory, outputBundleOptions.outputBaseDirectory, { overwrite: true }),
+    generateBundleYaml(outputBundleOptions, cwd),
+  ]);
+  return;
```
jamesdaniels, line 63: no need to explicitly return on a void

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -1,20 +1,87 @@
...
+    yamlStringify({
+      headers: [],
+      redirects: [],
+      rewrites: [],
+      runCommand: `node ${normalize(relative(cwd, outputBundleOptions.serverFilePath))}`,
+      neededDirs: [normalize(relative(cwd, outputBundleOptions.outputDirectory))],
+      staticAssets: [normalize(relative(cwd, outputBundleOptions.browserDirectory))],
+    }),
+  );
+  return;
```
jamesdaniels, line 82: no need to explicitly return

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -46,7 +49,14 @@ export async function loadConfig(cwd: string) {
 
   const options = await architectHost.getOptionsForTarget(buildTarget);
   if (!options) throw new Error("Not able to find options for build target.");
-  return options;
+
+  // options has to be of type ApplicationBuilderOptions when the builder is an application builder
+  const applicationBuilderOptions: ApplicationBuilderOptions = Object.assign(
```
jamesdaniels, line 54: Can we cooerce the type without an object assign? also you can return this object immediately w/o the intermediary const

jamesdaniels (changes requested):

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -1,20 +1,87 @@
...
     process.on("exit", (code) => {
       if (code === 0) return resolve();
       reject();
     });
   });
 
-const config = await loadConfig(process.cwd());
+// move the base output directory, which contains the server and browser bundle directory, and prerendered routes
+// as well as generating bundle.yaml
+async function generateOutputDirectory(
```
Yuangwang, line 53: Could we move this function + populateOutputBundleOptions + any other dependent functions into utils.js? It will help facilitate the unit tests

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -1,20 +1,87 @@
...
+      staticAssets: [normalize(relative(cwd, outputBundleOptions.browserDirectory))],
+    }),
+  );
+  return;
+}
 
-// TODO do all the things
-console.log({ config });
+const outputBundleOptions = populateOutputBundleOptions();
+await build().catch(() => process.exit(1));
```
Yuangwang, line 86: Whats the intention of catching and then doing process.exit?

Yuangwang (changes requested):
Could you add unit tests as well? similar to https://github.com/FirebaseExtended/firebase-framework-tools/blob/main/packages/%40apphosting/adapter-nextjs/src/bin/build.spec.ts

packages/@apphosting/adapter-angular/src/bin/build.ts
sjjj986: The 'spawn' event will fire regardless of whether an error occurs within the spawned process. So in case reject(), we can catch and exit with 1

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -1,20 +1,16 @@
...
-const build = (cwd = process.cwd()) =>
-  new Promise<void>((resolve, reject) => {
-    // TODO warn if the build script contains anything other than `ng build`
-    const process = spawn("npm", ["run", "build"], { cwd, shell: true, stdio: "inherit" });
-    process.on("exit", (code) => {
-      if (code === 0) return resolve();
-      reject();
-    });
-  });
+const cwd = process.cwd();
```
leoortizz, line 9: is this necessary?

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -1,20 +1,16 @@
...
 
-const config = await loadConfig(process.cwd());
+// read Application Builder config
```
leoortizz, line 11: probably better to move this comment to describe the function

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -1,10 +1,17 @@
...
 
 // fs-extra is CJS, readJson can't be imported using shorthand
-export const { readJson } = fsExtra;
+export const { writeFile, move, readJson } = fsExtra;
 
-export async function loadConfig(cwd: string) {
+export type { ApplicationBuilderOptions };
+
+export async function loadConfig(cwd: string): Promise<ApplicationBuilderOptions> {
   // dynamically load NextJS so this can be used in an NPX context
```
leoortizz, line 15: not from this PR but this should say Angular instead

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -46,7 +53,81 @@ export async function loadConfig(cwd: string) {
...
+    process.env.NG_BUILD_LOGS_JSON = "1";
+    const childProcess = spawn("npm", ["run", "build"], { cwd, shell: true, stdio: "inherit" });
+    childProcess.on("exit", (code) => {
+      if (code === 0) return resolve();
+      reject();
+    });
+  });
+
+// move the base output directory, which contains the server and browser bundle directory, and prerendered routes
+// as well as generating bundle.yaml
```
leoortizz, line 102: better to use multi-line comments instead for better support in the code editor for all functions

package.json
```diff
@@ -25,6 +26,7 @@
     "mocha": "^10.2.0",
     "prettier": "^3.0.3",
     "ts-mocha": "^10.0.0",
-    "typescript": "^5.2.0"
+    "typescript": "^5.2.0",
+    "zod": "^3.22.4"
```
leoortizz, line 30: I don't see `zod` being used here?

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: removed this import to use build manifest

package.json
sjjj986: added zod validation

packages/@apphosting/adapter-angular/src/bin/build.ts
sjjj986: done

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: done

packages/@apphosting/adapter-angular/src/bin/build.ts
sjjj986: done

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: added zod validation

packages/@apphosting/adapter-angular/src/bin/build.ts
sjjj986: done

packages/@apphosting/adapter-angular/src/bin/build.ts
sjjj986: added manifest support

sjjj986:
Removed manually loading angular application builder options & normalizing options
Use build manifest from stdout, verify its format with zod, and then extract the output paths.
Errors and warnings from the manifest will be logged.

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -32,21 +40,114 @@ export async function loadConfig(cwd: string) {
...
+            if (manifest["warnings"].length > 0) {
+              // warnings when extracting manifest
+              manifest.warnings.forEach((warning) => {
+                logger.info(warning);
+              });
+            }
+            if (manifest["outputPaths"]) {
+              outputPathOptions = populateOutputBundleOptions(manifest["outputPaths"]);
+            } else {
+              throw new Error("Could not find output paths from the build manifest.");
```
Yuangwang, line 105: In the case where this  error is thrown and then caught in line 108, what happens to outputPathOptions? Is it {} still? Does that work still?

packages/@apphosting/adapter-angular/src/interface.ts
```diff
@@ -0,0 +1,50 @@
...
+export const buildManifestSchema = z.object({
+  errors: z.array(z.string()),
+  warnings: z.array(z.string()),
+  outputPaths: z
+    .object({
+      root: url,
+      server: z.optional(url),
+      browser: url,
+    })
+    .optional(),
```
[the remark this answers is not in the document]
sjjj986: should not be optional, thanks

.eslintrc.cjs
```diff
@@ -13,6 +13,7 @@ module.exports = {
   ],
   rules: {
     "jsdoc/newline-after-description": "off",
+    "@typescript-eslint/no-explicit-any": "off", // TODO(sijin): remove later
```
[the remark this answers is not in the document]
sjjj986: to do the zod custom url type I have to use type any, couldn't figure out how to make it work without turning off "no-explicit-any"

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -32,21 +40,114 @@ export async function loadConfig(cwd: string) {
...
+            }
+          }
+        } catch (err) {
+          logger.error("Build manifest is not of expected structure: " + err);
+        }
+      });
+    }
+    childProcess.on("exit", (code) => {
+      if (code === 0) return resolve(outputPathOptions);
```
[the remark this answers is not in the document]
sjjj986: the resolve here is the resolve from resolve/reject, and should return Promise<OutputPathOptions>

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: Outputpaths should always be there, so I fixed this part. This manifest should be validated by zod and outputPathOptions should not be empty

packages/@apphosting/adapter-angular/src/utils.ts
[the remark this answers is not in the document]
Yuangwang: Ahh I got confused with path.resolve

Yuangwang (approved):

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -32,21 +40,114 @@ export async function loadConfig(cwd: string) {
...
+    serverFilePath: resolve(outputBundleDir, "dist", serverRelativePath, "server.mjs"),
+    browserDirectory: resolve(outputBundleDir, "dist", browserRelativePath),
   };
+}
 
-  const options = await architectHost.getOptionsForTarget(buildTarget);
-  if (!options) throw new Error("Not able to find options for build target.");
-  return options;
+// Run build command
+export const build = (cwd = process.cwd()) =>
```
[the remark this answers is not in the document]
Yuangwang: Synced offline, this should be ok to leave as is for now

.eslintrc.cjs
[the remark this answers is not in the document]
jamesdaniels: You can add a eslint-disable-next-line comment above the line in question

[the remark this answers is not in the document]
leoortizz: btw `unknown` is an alternative to `any` that doesn't require changing the ESLint rules

package.json
jamesdaniels, line 30: also move to package.json in packages/@apphosting/adapter-angular

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -32,21 +40,113 @@ export async function loadConfig(cwd: string) {
...
+              });
+            }
+            outputPathOptions = populateOutputBundleOptions(manifest["outputPaths"]);
+          }
+        } catch (error) {
+          throw new Error("Build manifest is not of expected structure: " + error);
+        }
+      });
+    } else {
+      throw new Error("Unable to locate build manifest with output paths.");
```
jamesdaniels, line 109: let's not throw here, ignore since the SSG / other build scripts might be emitting to stdout

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -32,21 +40,113 @@ export async function loadConfig(cwd: string) {
...
+export const build = (cwd = process.cwd()) =>
+  new Promise<OutputPathOptions>((resolve, reject) => {
+    // enable JSON build logs for application builder
+    process.env.NG_BUILD_LOGS_JSON = "1";
+    const childProcess = spawn("npm", ["run", "build"], {
+      cwd,
+      shell: true,
+      stdio: ["inherit", "pipe", "pipe"],
+    });
+    let outputPathOptions = {} as OutputPathOptions;
```
jamesdaniels, line 80: instead, let's make this `let outputPathOptions: OutputPathOptions | undefined;` and on exit, throw is outputPathOptions is still undefined

jamesdaniels (approved):
Added some nits, if you have time feel free to address otherwise we can circle back

PR #145, opened 2024-02-21
simple server for angular CSR

Yuangwang:


jamesdaniels (approved):
LGTM, especially since this will not be a permanent fixture in the codebase—eventually we'll support static only.

PR #147, opened 2024-02-29
Add output validations

Yuangwang:
Add validations check that the output bundle generated is as expected

### DIFF
--- a/packages/@apphosting/adapter-angular/src/utils.ts
+++ b/packages/@apphosting/adapter-angular/src/utils.ts
@@ -211,7 +211,8 @@ export async function validateOutputDirectory(
   if (
     !(await fsExtra.exists(outputBundleOptions.outputDirectory)) ||
     !(await fsExtra.exists(outputBundleOptions.browserDirectory)) ||
-    !(await fsExtra.exists(outputBundleOptions.serverFilePath))
+    !(await fsExtra.exists(outputBundleOptions.serverFilePath)) ||
+    !(await fsExtra.exists(outputBundleOptions.bundleYamlPath))
   ) {
     throw new Error("Output directory is not of expected structure");
   }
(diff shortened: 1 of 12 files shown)


packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -161,39 +166,52 @@ function extractManifestOutput(output: string): string {
...
+}
+
+// Validate output directory includes all necessary parts
+export async function validateOutputDirectory(
+  outputBundleOptions: OutputBundleOptions,
+): Promise<void> {
+  if (
+    !(await fsExtra.exists(outputBundleOptions.outputDirectory)) ||
+    !(await fsExtra.exists(outputBundleOptions.browserDirectory)) ||
+    !(await fsExtra.exists(outputBundleOptions.serverFilePath))
```
sjjj986, line 211: Should we also verify the existence of bundleYamlPath

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -109,3 +109,15 @@ async function generateBundleYaml(
...
   return;
 }
+
+// Validate output directory includes all necessary parts
+export async function validateOutputDirectory(
+  outputBundleOptions: OutputBundleOptions,
+): Promise<void> {
+  if (
+    !(await fsExtra.exists(outputBundleOptions.outputDirectory)) ||
+    !(await fsExtra.exists(outputBundleOptions.serverFilePath))
```
sjjj986, line 119: Same here should we check bundle.yaml exists?

packages/@apphosting/adapter-angular/src/utils.ts
Yuangwang: Yes good call I added these checks

packages/@apphosting/adapter-nextjs/src/utils.ts
Yuangwang: Added

sjjj986 (approved):

PR #151, opened 2024-03-20
Support monorepo builds in Angular adapter

blidd-google:
Currently, the Angular adapter is unable to specify a custom build command to run the Angular application builder. It also makes certain assumptions that no longer hold true in monorepo contexts. We adjust those conditions to enable builds of Angular projects using monorepo tooling.

### DIFF
--- a/packages/@apphosting/adapter-angular/src/bin/build.ts
+++ b/packages/@apphosting/adapter-angular/src/bin/build.ts
@@ -1,5 +1,11 @@
 #! /usr/bin/env node
-import { checkBuildConditions, build, generateOutputDirectory, DEFAULT_COMMAND } from "../utils.js";
+import {
+  build,
+  generateOutputDirectory,
+  DEFAULT_COMMAND,
+  checkStandaloneBuildConditions,
+  checkMonorepoBuildConditions,
+} from "../utils.js";
 
 const root = process.cwd();
 
@@ -7,8 +13,10 @@ const root = process.cwd();
 let projectRoot = root;
 if (process.env.MONOREPO_PROJECT && process.env.FIREBASE_APP_DIRECTORY) {
   projectRoot = projectRoot.concat("/", process.env.FIREBASE_APP_DIRECTORY);
+  const builder = process.env.MONOREPO_BUILDER || "";
+  checkMonorepoBuildConditions(builder);
 } else {
-  await checkBuildConditions(projectRoot);
+  await checkStandaloneBuildConditions(projectRoot);
 }
 
 // Determine which build runner to use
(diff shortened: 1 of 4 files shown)


packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -68,56 +71,76 @@ export function populateOutputBundleOptions(outputPaths: OutputPaths): OutputPat
...
+      }
     });
   });
 
-/* 
-Move the base output directory, which contains the server and browser bundle directory, and prerendered routes
-as well as generating bundle.yaml.
+/**
+ * Extracts the build manifest from the build command's console output.
+ * N.B. Unfortunately, there is currently no consistent way to suppress extraneous default output from the task
```
Yuangwang, line 126: ooc what is nx sending to std out?

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -1,9 +1,21 @@
...
-await checkBuildConditions(cwd);
+// Determine root of project to build
+let projectRoot = root;
+if (process.env.MONOREPO_PROJECT && process.env.FIREBASE_APP_DIRECTORY) {
+  projectRoot = projectRoot.concat("/", process.env.FIREBASE_APP_DIRECTORY);
+} else {
+  await checkBuildConditions(projectRoot);
+}
 
-const outputBundleOptions = await build().catch(() => process.exit(1));
```
Yuangwang, line 8: Why was the .catch(() => process.exit(1)) removed here?

blidd-google: Oh good catch - I removed this while debugging because the program would silently exit without providing the specific error message. If we are intentionally hiding the error, I can add this back in.

packages/@apphosting/adapter-angular/src/utils.ts
blidd-google: Looks like something like this:
Screenshot 2024-03-22 at 4 54 21 PM

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -1,9 +1,21 @@
...
+const root = process.cwd();
 
-await checkBuildConditions(cwd);
+// Determine root of project to build
+let projectRoot = root;
+if (process.env.MONOREPO_PROJECT && process.env.FIREBASE_APP_DIRECTORY) {
+  projectRoot = projectRoot.concat("/", process.env.FIREBASE_APP_DIRECTORY);
+} else {
+  await checkBuildConditions(projectRoot);
+}
```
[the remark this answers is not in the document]
blidd-google: Ah yes good catch, the Nx buildpack will read what builder is being used, then pass an env var MONOREPO_BUILDER to the Angular adapter.

packages/@apphosting/adapter-angular/src/bin/build.ts
Yuangwang: Ah gotcha, @sjjj986 do you remember why we needed this catch? Do you think its safe to remove?

sjjj986: looks good to me

sjjj986 (approved):

Yuangwang (approved):

PR #153, opened 2024-03-25
Fixed import to work with pnpm

sjjj986:
Fixed import to use require.resolve to work with pnpm file structure. Tested all three package managers work fine now.

### DIFF
--- a/packages/@apphosting/adapter-angular/src/utils.ts
+++ b/packages/@apphosting/adapter-angular/src/utils.ts
@@ -6,10 +6,12 @@ import { spawn } from "child_process";
 import { resolve, normalize, relative, dirname, join } from "path";
 import { stringify as yamlStringify } from "yaml";
 import { OutputPathOptions, OutputPaths, buildManifestSchema, ValidManifest } from "./interface.js";
+import { createRequire } from 'node:module';
 
 // fs-extra is CJS, readJson can't be imported using shorthand
 export const { writeFile, move, readJson, mkdir, copyFile } = fsExtra;
 
+const require = createRequire(import.meta.url);
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 const SIMPLE_SERVER_FILE_PATH = join(__dirname, "simple-server", "bundled_server.mjs");
@@ -21,12 +23,13 @@ const SIMPLE_SERVER_FILE_PATH = join(__dirname, "simple-server", "bundled_server
 */
 export async function checkBuildConditions(cwd: string): Promise<void> {
   // dynamically load Angular so this can be used in an NPX context
-  const { NodeJsAsyncHost }: typeof import("@angular-devkit/core/node") = await import(
-    `${cwd}/node_modules/@angular-devkit/core/node/index.js`
-  );
+  const { NodeJsAsyncHost }: typeof import("@angular-devkit/core/node") = await import (
+    require.resolve('@angular-devkit/core/node/index.js')
+    );
+
   const { workspaces }: typeof import("@angular-devkit/core") = await import(
-    `${cwd}/node_modules/@angular-devkit/core/src/index.js`
-  );
+    require.resolve('@angular-devkit/core/src/index.js')
+    );
 
   const host = workspaces.createWorkspaceHost(new NodeJsAsyncHost());
   const { workspace } = await workspaces.readWorkspace(cwd, host);


packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -22,10 +24,11 @@ const SIMPLE_SERVER_FILE_PATH = join(__dirname, "simple-server", "bundled_server
 export async function checkBuildConditions(cwd: string): Promise<void> {
   // dynamically load Angular so this can be used in an NPX context
   const { NodeJsAsyncHost }: typeof import("@angular-devkit/core/node") = await import(
-    `${cwd}/node_modules/@angular-devkit/core/node/index.js`
+    require.resolve("@angular-devkit/core/node/index.js")
```
jamesdaniels, line 27: You can use the cwd in the second argument, require.resolve(..., { paths: [cwd] }) this will help when invoked from NPX.

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -22,10 +24,11 @@ const SIMPLE_SERVER_FILE_PATH = join(__dirname, "simple-server", "bundled_server
 export async function checkBuildConditions(cwd: string): Promise<void> {
   // dynamically load Angular so this can be used in an NPX context
   const { NodeJsAsyncHost }: typeof import("@angular-devkit/core/node") = await import(
-    `${cwd}/node_modules/@angular-devkit/core/node/index.js`
+    require.resolve("@angular-devkit/core/node/index.js", { paths: [cwd] })
```
Yuangwang, line 27: Why do we need the paths:[cwd]?

If we remove this I think it will let us support yarn pnp once the Angular build issue is fixed

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -7,15 +7,17 @@ import { stringify as yamlStringify } from "yaml";
...
 import type { RoutesManifest } from "./interfaces.js";
 // fs-extra is CJS, readJson can't be imported using shorthand
 export const { move, exists, writeFile, readJson } = fsExtra;
 
 export async function loadConfig(cwd: string) {
+  const require = createRequire(import.meta.url);
   // dynamically load NextJS so this can be used in an NPX context
   const { default: nextServerConfig }: { default: typeof import("next/dist/server/config.js") } =
-    await import(`${cwd}/node_modules/next/dist/server/config.js`);
+    await import(require.resolve("next/dist/server/config.js", { paths: [cwd] }));
```
Yuangwang, line 20: I have a pr out for fixing yarn pnp in nextjs and it includes this change + a little extra to get yarn pnp working, https://github.com/FirebaseExtended/firebase-framework-tools/pull/156/files.

Could you revert these changes and we work off of that pr going forward?

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: @jamesdaniels James could you explain why only with cwd, npx would be supported? And if it's safe to remove this so that we can support yarn pnp?

jamesdaniels: So paths is a hint telling NPM to resolve in the specified directories. This is important incase this script is called via NPX, globally installed, or if the developer if using a version manager like Volta—as we'd want these dependencies to be resolved specifically for the project at hand. 

Not a big deal for the buildpack but this guards the web frameworks migration, integration testing, and emulation paths in the future.

Yuangwang: I think we run into a problem with yarn plugnplay if we restrict npm to only checking in cwd. Yarn pnp (https://yarnpkg.com/features/pnp) doesn't use node_modules but uses a cache instead where it installs the dependencies.

For example some app using yarn pnp might have the @angular-devkit/core module resolve to ../../.yarn/cache/@angular-devkit/core. If we keep "paths: [cwd]" then it will never find this directory.

The location of this cache directory is also optionally configurable via a field in .yarnrc.yml so its also not as simple as just adding another path that we know will contain this .yarn cache.

Going forward if we want to support yarn pnp we will need to either drop this cwd or add in some custom logic to find the .yarn dir to add into paths as well. I would lean towards dropping the cwd here for now and revisiting this if it becomes a bigger issue in the future, I'm not the biggest fan of adding a good amount of custom logic to support yarn pnp here specifically. Open to other thoughts or ideas as well. wdyt @jamesdaniels?

jamesdaniels: Shouldn't corepack handle that for us? Are we enabling it in the buildpack if package manager != npm

Yuangwang: Let's get these changes in now imo. We should probably get corepack to handle this for us in the long term but we probably want to have pnpm working for now.

packages/@apphosting/adapter-nextjs/src/utils.ts
Yuangwang: Actually lets get your changes in instead of my new pr. I think we should get pnpm working and maybe try to get everything setup with corepack as a longer term solution

Yuangwang (approved):

packages/@apphosting/adapter-nextjs/src/utils.ts
sjjj986: ok sounds good, let me retest and submit this one. Will take a look at corepack later

PR #155, opened 2024-03-27
Add cache control header

sjjj986:
Add cache control header to include maxAge=1 year for our Angular CSR simple server, so that it matches what Angular SSR server automatically configured

### DIFF
--- a/packages/@apphosting/adapter-angular/src/simple-server/server.mjs
+++ b/packages/@apphosting/adapter-angular/src/simple-server/server.mjs
@@ -4,8 +4,9 @@ import { fileURLToPath } from "url";
 const app = express();
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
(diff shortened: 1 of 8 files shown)


packages/@apphosting/adapter-angular/src/simple-server/server.mjs
```diff
@@ -4,8 +4,9 @@ import { fileURLToPath } from "url";
 const app = express();
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
+const oneYear = 31_557_600_000;
```
jamesdaniels, line 7: TIL express maxAge is in milliseconds but let's stick with traditional seconds-calendar year calculation 31_536_000_000

sjjj986: Done.

Yuangwang (approved):

PR #156, opened 2024-03-27
Fix nextjs yarn pnp integration

Yuangwang:
Currently the nextjs adapter doesn't work if the app is built using yarn plug n play, this fixes that.

### DIFF
--- a/packages/@apphosting/adapter-nextjs/src/utils.ts
+++ b/packages/@apphosting/adapter-nextjs/src/utils.ts
@@ -49,7 +49,7 @@ export async function build(cwd: string): Promise<void> {
   process.env.NEXT_PRIVATE_STANDALONE = "true";
   // Opt-out sending telemetry to Vercel
   process.env.NEXT_TELEMETRY_DISABLED = "1";
-  if (!(await exists(join(cwd, "node_modules")))) {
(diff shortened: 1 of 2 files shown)


packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -42,12 +44,16 @@ export function populateOutputBundleOptions(cwd: string): OutputBundleOptions {
...
 
 // Run build command
-export function build(cwd: string): void {
+export async function build(cwd: string): Promise<void> {
   // Set standalone mode
   process.env.NEXT_PRIVATE_STANDALONE = "true";
   // Opt-out sending telemetry to Vercel
   process.env.NEXT_TELEMETRY_DISABLED = "1";
-  spawnSync("npm", ["run", "build"], { cwd, shell: true, stdio: "inherit" });
+  if (!(await exists(join(cwd, "node_modules")))) {
```
jamesdaniels, line 52: Could we add a more robust test here? Maybe check for a yarn lock? Alt could we expose the package manager via an environment variable in the buildpack?

Yuangwang: Good call, I changed this to check for yarn.lock.

PR #158, opened 2024-04-04
Set correct port to listen to in Angular buildpack

sjjj986:
b/332551808

### DIFF
--- a/packages/@apphosting/adapter-angular/src/utils.ts
+++ b/packages/@apphosting/adapter-angular/src/utils.ts
@@ -183,6 +183,7 @@ async function generateBundleYaml(
       headers: [],
       redirects: [],
       rewrites: [],
+      // this fix is needed for Angular version 17.3.2
       runCommand: `SSR_PORT=$PORT node ${normalize(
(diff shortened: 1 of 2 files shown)


packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -183,7 +183,9 @@ async function generateBundleYaml(
       headers: [],
       redirects: [],
       rewrites: [],
-      runCommand: `node ${normalize(relative(cwd, outputPathOptions.serverFilePath))}`,
+      runCommand: `SSR_PORT=$PORT node ${normalize(
```
Yuangwang, line 186: Can you add a comment that note what version this fix is needed for, we many want to have a separate versioned adapter in the future

sjjj986: added a comment below

sjjj986: My bad, just added!

Yuangwang (approved):

PR #159, opened 2024-04-06
fix: sveltekit server endpoint buffer response

m3tasploit:
Fix for corrupt images when responding with image/jpeg or any binary data from sveltekit Api endpoint


jamesdaniels:
@m3tasploit LGTM. Q, does the text content-type repsonse work ok if you always respond with a buffer? Best to avoid conditionals when possible

m3tasploit:
Oh, sorry i forgot that case, I'll try different content types and update code of necessary

m3tasploit:
> @m3tasploit LGTM. Q, does the text content-type repsonse work ok if you always respond with a buffer? Best to avoid conditionals when possible

I've added check for commonly used text content-types. I'm not sure if this is the correct way to go.

leoortizz:
Hey @m3tasploit, thanks for creating this PR. I did a few tests and found that if we always respond with a buffer it also works, therefore avoiding conditionals. Implemented in #205.

leoortizz:
Merged in https://github.com/FirebaseExtended/firebase-framework-tools/pull/205

PR #160, opened 2024-04-10
Support monorepo builds in Next.js adapter

blidd-google:
Currently, the Next.js adapter is unable to specify a custom build command to run the Angular application builder. It also makes certain assumptions that no longer hold true in monorepo contexts. We adjust those conditions to enable builds of Next.js projects using monorepo tooling.


packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -1,21 +1,27 @@
...
-import type { RoutesManifest } from "./interfaces.js";
 // fs-extra is CJS, readJson can't be imported using shorthand
 export const { move, exists, writeFile, readJson } = fsExtra;
 
-export async function loadConfig(cwd: string) {
+export const DEFAULT_COMMAND = "npm";
+
+export async function loadConfig(cwd: string): Promise<NextConfigComplete> {
   // dynamically load NextJS so this can be used in an NPX context
+  const require = createRequire(import.meta.url);
```
Yuangwang, line 20: Why do we need this change? We might have discussed this but I forgot the reason

packages/@apphosting/adapter-nextjs/src/bin/build.spec.ts
```diff
@@ -17,6 +18,13 @@ describe("build commands", () => {
       outputStaticDirectory: path.join(tmpDir, ".apphosting/.next/static"),
       serverFilePath: path.join(tmpDir, ".apphosting/server.js"),
     };
+    outputBundleMonorepoOptions = {
```
tonyjhuang, line 21: nit I would inline this, see go/tott/643

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -1,21 +1,27 @@
...
+import { ROUTES_MANIFEST } from "./constants.js";
+import { OutputBundleOptions, RoutesManifest } from "./interfaces.js";
+import { NextConfigComplete } from "next/dist/server/config-shared.js";
 
-import type { RoutesManifest } from "./interfaces.js";
 // fs-extra is CJS, readJson can't be imported using shorthand
 export const { move, exists, writeFile, readJson } = fsExtra;
 
-export async function loadConfig(cwd: string) {
+export const DEFAULT_COMMAND = "npm";
```
tonyjhuang, line 16: let's make sure we have java docs for all public/exported functions and especially constants

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -30,42 +36,46 @@ export const isMain = (meta: ImportMeta) => {
   return process.argv[1] === fileURLToPath(meta.url);
 };
 
-export function populateOutputBundleOptions(cwd: string): OutputBundleOptions {
-  const outputBundleDir = join(cwd, ".apphosting");
+export function populateOutputBundleOptions(rootDir: string, appDir: string): OutputBundleOptions {
+  const outputBundleDir = join(rootDir, ".apphosting");
```
tonyjhuang, line 40: can we make sure to add some javadocs for the params here too? What is a rootdir, what is an appdir?

In particular, I'm surprised that we're looking to write .apphosting to the rootdir. It would seem more appropriate to write it to the same levels as the application directory.

If in the future we support building multiple backends in one build process, writing it to the rootdir will cause subsequent builds to stomp on previous build artifacts

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -30,42 +36,46 @@ export const isMain = (meta: ImportMeta) => {
...
+  // In monorepo setups, the standalone directory structure will mirror the structure of the monorepo.
```
tonyjhuang, line 41: Is this only the case for monorepo setups?

What about the following project setup where `appdir = /nested/app/src`?
```
/
| nested
- | app
  - | src
```

Perhaps we should make this comment a bit more general to any non-"/" appdir setups

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -4,16 +4,27 @@ import {
...
+
+let projectRoot = root;
+if (process.env.MONOREPO_PROJECT && process.env.FIREBASE_APP_DIRECTORY) {
+  projectRoot = projectRoot.concat("/", process.env.FIREBASE_APP_DIRECTORY);
+}
+
+// Determine which build runner to use
+let cmd = DEFAULT_COMMAND;
+if (process.env.MONOREPO_COMMAND) {
+  cmd = process.env.MONOREPO_COMMAND;
```
tonyjhuang, line 21: Is this generalizable to all monorepos?

Seems like `turborepo` uses `turborepo run ...` which seems appropriate, lerna uses `npx lerna run ...`

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -4,16 +4,27 @@ import {
...
+  DEFAULT_COMMAND,
 } from "../utils.js";
-
 import { join } from "path";
 
-const cwd = process.cwd();
+const root = process.cwd();
+
+let projectRoot = root;
+if (process.env.MONOREPO_PROJECT && process.env.FIREBASE_APP_DIRECTORY) {
```
tonyjhuang, line 14: OOC why do we need the MONOREPO_PROJECT check? Shouldn't we always respect FIREBASE_APP_DIRECTORY?

tonyjhuang (changes requested):

tonyjhuang:
Overall, this seems like a reasonable approach and I see now that this follows the pattern used by the Angular adapter as well.

My high level feedback is: Any env vars that we introduce as inputs into the build adapter should be treated as a public API for other framework (adapter) authors down the road. We should make sure that this API is 1) thoroughly documented, and 2) follows our usual standards of API design including go/easy-to-use-and-hard-to-misuse (also see go/tott/523).

In particular, I would recommend re-examining our current set of env vars. As of right now, we have:
* `process.env.MONOREPO_PROJECT`
* `process.env.FIREBASE_APP_DIRECTORY`
* `process.env.MONOREPO_BUILDER`
* `process.env.MONOREPO_COMMAND`

It seems like some (MONOREPO_BUILDER) are not always used, while others (MONOREPO_PROJECT) can probably be removed all together. It also seems like a bug that FIREBASE_APP_DIRECTORY is only ever respected in MONOREPO contexts. IMO we should either rename it to MONOREPO_APP_DIRECTORY or always respect it.

tonyjhuang:
I wonder if our API could be simplified down to: 

`process.env.MONOREPO_PROVIDER = "nx"`?

This gives (community) adapter authors the most flexibility as opposed to having to handle our buildpack-specific logic such as `MONOREPO_COMMAND`, and we can have a shared util for example:

`getBuildCommand(monorepoProvider) => providerToBuildCommand[monorepoProvider];`

blidd-google:
> It seems like some (MONOREPO_BUILDER) are not always used

MONOREPO_BUILDER is not used in the Next.js buildpack, but is necessary for the Angular buildpack. The Nx buildpack is responsible for reading project config files (i.e. project.json), which is how we determine which application builder Angular is using in an Nx monorepo (in non-Nx environments we would read the angular.json, but there is no angular.json for Nx + Angular projects). We need some way to pass the information of which app builder is being used. The alternative would be to  have the adapter read Nx project config files.

I do agree that MONOREPO_BUILDER is optional in many circumstances, however. This should be reflected in documentation.

> while others (MONOREPO_PROJECT) can probably be removed all together

In the case of Nx, this is true - the MONOREPO_PROJECT is not strictly necessary because we can simply run `nx run build` from the context of the app directory and Nx will only build that app. Nx also supports `nx run [project]:build` to build a specific project. However, `turbo run build` by default builds every project regardless of the cwd context, and requires a `--filter` flag to run a build task for a single project ([docs](https://turbo.build/repo/docs/core-concepts/monorepos/filtering)). So technically in the adapters, we should be using MONOREPO_PROJECT, but I haven't made the change yet.

> It also seems like a bug that FIREBASE_APP_DIRECTORY is only ever respected in MONOREPO contexts. IMO we should either rename it to MONOREPO_APP_DIRECTORY or always respect it.

We should probably always respect it to support setups where the repo root != the app root.

packages/@apphosting/adapter-nextjs/src/bin/build.ts
blidd-google: Yes we should always be respecting FIREBASE_APP_DIRECTORY.

packages/@apphosting/adapter-nextjs/src/bin/build.ts
blidd-google: Yes, they mirror the syntax of npm scripts (`npm run build`, `nx run build`, `turbo run build`, etc.)

blidd-google: However some have specific syntax, e.g. Nx allows specifying `nx run [project]:build:[environment]`.

packages/@apphosting/adapter-nextjs/src/utils.ts
blidd-google: Yes it should support the project setup you described, although I should note that there are some rough edges to smooth out to support these non-monorepo nested setups.

packages/@apphosting/adapter-nextjs/src/utils.ts
blidd-google: Yeah we discussed this a few weeks ago and my memory was also a bit fuzzy, but we need this in order to locate the package in the root node_modules folder (which is relevant for monorepo contexts).

tonyjhuang: Is there any additional documentation we can link to in this line comment?

packages/@apphosting/adapter-nextjs/src/bin/build.ts
tonyjhuang: This seems like an argument for offloading the build command syntax to a shared library in the adapters

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -1,23 +1,31 @@
...
+import { OutputBundleOptions, RoutesManifest } from "./interfaces.js";
+import { NextConfigComplete } from "next/dist/server/config-shared.js";
 
-import type { RoutesManifest } from "./interfaces.js";
 // fs-extra is CJS, readJson can't be imported using shorthand
 export const { move, exists, writeFile, readJson } = fsExtra;
 
-export async function loadConfig(cwd: string) {
+// The default fallback command prefix to run a build.
+export const DEFAULT_COMMAND = "npm";
```
tonyjhuang, line 17: Why npm?

tonyjhuang (commented):
Overall LGTM pending a few small nits.

My main ask following this CL is to see if we can encapsulate all of the env var wrangling in a single nx library that can be shareable across both the angular and the next adapters

blidd-google:
> Overall LGTM pending a few small nits.
> 
> My main ask following this CL is to see if we can encapsulate all of the env var wrangling in a single nx library that can be shareable across both the angular and the next adapters

Sounds good - do you mean a single monorepo library?

packages/@apphosting/adapter-nextjs/src/utils.ts
tonyjhuang: Sorry not to be too pedantic but this comment explains _what_ this does as opposed to _why_ we need it.

Can we have a comment here explaining the motivation behind this statement? eg "Intentionally use ESM module resolution here to support blah blah blah"

See [go/tott/563](https://goto.google.com/tott/563)

tonyjhuang: A good litmus test is: Would someone who's not you, a year from now, understand why we need this statement based on the line comment? Would they be able to gather sufficient context either from the code, the comment, or linked docs to own it going forward?

tonyjhuang (approved):

PR #161, opened 2024-04-11
Add a new "create-next-on-firebase" package

jhuleatt:
This package shells out to `@apphosting/create`, but specifies Next.js as the framework.

```
npm init next-on-firebase <directory>
```

will call

```
npx @apphosting/create --framework=nextjs <directory>
```


jhuleatt:
Looks like I'll need to update [`scripts/publish.js`](https://github.com/FirebaseExtended/firebase-framework-tools/blob/main/scripts/publish.js) as well - @jamesdaniels any advice on that?

jamesdaniels (approved):

PR #165, opened 2024-04-15
Make sure broken build breaks the presubmits

Yuangwang:
Make sure broken build breaks the presubmits and also push earlier server changes to the bundled server


packages/@apphosting/adapter-angular/src/simple-server/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "scripts": {
-      "build": "rollup -c --bundleConfigAsCjs"
+      "generate-server": "rollup -c --bundleConfigAsCjs"
```
sjjj986, line 3: what is this change for

Yuangwang: Previously the build was breaking because this was build (it was being picked up by the lerna build) but I reverted it and it seems to be working fine so I switched it back.

sjjj986 (approved):

PR #167, opened 2024-04-16
Update application builder error message

sjjj986:
Update application builder error message and point to angular.dev guide for upgrading to the application builder.


Yuangwang (approved):

sysia48:
Na początku było API z Gemini na strefadk.club

PR #170, opened 2024-04-22
Set up publishing App Hosting adapters to local registry for buildpacks testing

blidd-google:
During local development, we may want to iterate on early changes to a build adapter in the context of the buildpacks, without cutting a release to the @apphosting npm registry. This PR enables publishing build adapter prototypes to a local registry for development.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /packages/**/*/dist/
 /examples/
 /dist
+/packages/@apphosting/publish-dev/.local
 
 # NPM
 node_modules
--- a/packages/@apphosting/adapter-angular/package.json
+++ b/packages/@apphosting/adapter-angular/package.json
@@ -22,7 +22,9 @@
     "sideEffects": false,
     "scripts": {
         "build": "rm -rf dist && tsc && chmod +x ./dist/bin/*",
-        "test": "ts-mocha -p tsconfig.json src/**/*.spec.ts"
+        "test": "ts-mocha -p tsconfig.json src/**/*.spec.ts",
+        "registry:local": "npx verdaccio --config ../publish-dev/verdaccio-config.yaml",
+        "publish:local": "npm unpublish --@apphosting:registry=http://localhost:4873 --force && npm publish --@apphosting:registry=http://localhost:4873"
     },
     "exports": {
         ".": {
@@ -70,6 +72,7 @@
         "tmp": "*",
         "ts-mocha": "*",
         "ts-node": "*",
-        "typescript": "*"
+        "typescript": "*",
+        "verdaccio": "^5.30.3"
     }
 }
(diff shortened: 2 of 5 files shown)


packages/@apphosting/adapter-angular/package.json
```diff
@@ -22,7 +22,9 @@
     "sideEffects": false,
     "scripts": {
         "build": "rm -rf dist && tsc && chmod +x ./dist/bin/*",
-        "test": "ts-mocha -p tsconfig.json src/**/*.spec.ts"
+        "test": "ts-mocha -p tsconfig.json src/**/*.spec.ts",
+        "registry:local": "npx verdaccio --config ../publish-dev/verdaccio-config.yaml",
+        "publish:local": "npm unpublish --@apphosting:registry=http://localhost:4873 --force && npm publish --@apphosting:registry=http://localhost:4873"
```
Yuangwang, line 27: The unpublish causes some issues on the first run since there is nothing to unpublish, I get this error Not Found - GET http://localhost:4873/@apphosting%2fadapter-nextjs?write=true - no such package available

.gitignore
```diff
@@ -1,6 +1,7 @@
 /packages/**/*/dist/
 /examples/
 /dist
+/packages/@apphosting/publish-dev/.local
```
Yuangwang, line 5: I notice that running npm adduser --registry http://localhost:4873 generates packages/@apphosting/publish-dev/htpasswd which holds the credentials I entered. We probably want that in the .gitignore too to prevent any accidental credential leaks.

packages/@apphosting/adapter-nextjs/package.json
```diff
@@ -22,7 +22,9 @@
   "sideEffects": false,
   "scripts": {
     "build": "rm -rf dist && tsc && chmod +x ./dist/bin/*",
-    "test": "ts-mocha -p tsconfig.json src/**/*.spec.ts"
+    "test": "ts-mocha -p tsconfig.json src/**/*.spec.ts",
+    "registry:local": "npx verdaccio --config ../publish-dev/verdaccio-config.yaml",
+    "publish:local": "npm unpublish --@apphosting:registry=http://localhost:4873 --force && npm publish --@apphosting:registry=http://localhost:4873"
```
Yuangwang, line 27: Same comment as above

packages/@apphosting/publish-dev/verdaccio-config.yaml
```diff
@@ -0,0 +1,76 @@
+#
+# This is the default configuration file. It allows all users to do anything,
+# please read carefully the documentation and best practices to
+# improve security.
```
Yuangwang, line 4: We should probably remove this bit about the default configuration file since it looks like you've made some modifications

packages/@apphosting/adapter-angular/package.json
sjjj986: also running into this issue http://localhost:4873/@apphosting%2fadapter-angular?write=true

PR #176, opened 2024-05-07
Copy all files in nextjs adapter

Yuangwang:
Copies over all the files to .apphosting in the nextjs build adapter. This fixes issues where a library relies on reading a file that previously was not copied over


blidd-google (approved):

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -96,23 +98,28 @@ export async function generateOutputDirectory(
...
 
-// move public directory to apphosting output public directory
-async function movePublicDirectory(
-  publicDirectory: string,
-  appHostingPublicDirectory: string,
-): Promise<void> {
-  const publicDirectoryExists = await exists(publicDirectory);
-  if (!publicDirectoryExists) return;
-  await move(publicDirectory, appHostingPublicDirectory, { overwrite: true });
+// Move all files and directories directory to apphosting output directory.
```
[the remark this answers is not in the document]
blidd-google: Just to make sure I understand - we are simply copying over everything from the application directory now? These files will stay on the server right (since they're not being specified as static or public assets)?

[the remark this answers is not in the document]
Yuangwang: Yeah all the application dir files are being copied over now to be used on the server.

sjjj986 (approved):

packages/@apphosting/adapter-nextjs/src/interfaces.ts
```diff
@@ -76,6 +76,7 @@ export interface RoutesManifest {
 export interface OutputBundleOptions {
   bundleYamlPath: string;
   outputDirectory: string;
+  outputBundleAppDir: string;
```
tonyjhuang, line 79: I think it's time to document this interface :) Can you add jsdocs for each field to explain what they are? Can you also do an audit to make sure the field names are as unambiguous as possible?

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -96,23 +98,28 @@ export async function generateOutputDirectory(
...
-): Promise<void> {
-  const publicDirectoryExists = await exists(publicDirectory);
-  if (!publicDirectoryExists) return;
-  await move(publicDirectory, appHostingPublicDirectory, { overwrite: true });
+// Move all files and directories to apphosting output directory.
+// Files are skipped if there is already a file with the same name in the output directory
+async function moveResources(appDir: string, outputBundleAppDir: string): Promise<void> {
+  const appDirExists = await exists(appDir);
+  if (!appDirExists) return;
+  const dirsToMove = await readdir(appDir);
```
tonyjhuang, line 114: Are these results all guaranteed to be directories?

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -96,23 +98,28 @@ export async function generateOutputDirectory(
...
+  for (const dir of dirsToMove) {
+    if (
+      join(appDir, dir) !== outputBundleAppDir &&
```
tonyjhuang, line 117: This makes my head hurt :( Can we simplify these booleans

go/tott/685, go/tott/705

Yuangwang: Extracted out the booleans to be more clear

packages/@apphosting/adapter-nextjs/src/utils.ts
Yuangwang: good catch, they can be files or dirs. I renamed the variable to account for this

packages/@apphosting/adapter-nextjs/src/interfaces.ts
Yuangwang: Added docs and renamed some variables to make things more clear

packages/@apphosting/adapter-nextjs/src/interfaces.ts
```diff
@@ -74,9 +74,29 @@ export interface RoutesManifest {
 
 // The output bundle options are specified here
 export interface OutputBundleOptions {
+  /**
+   * bundleYamlPath: Path to the generated bundle.yaml file
```
tonyjhuang, line 78: You can remove the `<variable>:` prefix here. 

Take a look at go/tsjs-style#jsdoc-general-form for how to format single line jsdocs

Yuangwang: done

packages/@apphosting/adapter-nextjs/src/interfaces.ts
```diff
@@ -74,9 +74,29 @@ export interface RoutesManifest {
...
+   */
   bundleYamlPath: string;
-  outputDirectory: string;
+  /**
+   *  Path to where the base output directory (.apphosting) is generated
+   */
+  outputDirectoryBasePath: string;
+  /**
+   * Path to where the App's code sits in the output directory.
+   * This will only be different from outputDirectoryPath in Monorepo scenerios.
```
tonyjhuang, line 87: This may be a dumb question but are these monorepo scenarios obvious? Should we have an example?

packages/@apphosting/adapter-nextjs/src/interfaces.ts
```diff
@@ -74,9 +74,29 @@ export interface RoutesManifest {
 
 // The output bundle options are specified here
 export interface OutputBundleOptions {
+  /**
+   * Path to the generated bundle.yaml file
+   */
   bundleYamlPath: string;
-  outputDirectory: string;
+  /**
+   *  Path to where the base output directory (.apphosting) is generated
```
tonyjhuang, line 82: Is this output always called .apphosting?

Yuangwang: yeah its always .apphosting

packages/@apphosting/adapter-nextjs/src/interfaces.ts
```diff
@@ -74,9 +74,29 @@ export interface RoutesManifest {
...
+   */
+  outputDirectoryBasePath: string;
+  /**
+   * Path to where the App's code sits in the output directory.
+   * This will only be different from outputDirectoryPath in Monorepo scenerios.
+   */
+  outputDirectoryAppPath: string;
+  /**
+   * Path to the generated server.js file in the output directory
```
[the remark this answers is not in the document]
Yuangwang: done

packages/@apphosting/adapter-nextjs/src/interfaces.ts
Yuangwang: There's a comment that explains this more where this is set https://github.com/FirebaseExtended/firebase-framework-tools/blob/main/packages/%40apphosting/adapter-nextjs/src/utils.ts#L56. I also added an example to make this more clear here

PR #177, opened 2024-05-07
Run Nx + Angular builds from root context

blidd-google:
Builds that include configuration files at the root of the monorepo weren't being bundled into the generated artifacts because the adapter was running the monorepo build command from the project subdirectory context. We fix this by running build commands from the root and specifying the target project.

This PR also includes a change to read Nx project.json files to get and validate the Angular builder specified by the user. The Nx buildpack was previously responsible for reading this value from the project.json configuration file; however, to support use cases in which `FIREBASE_APP_DIRECTORY` is not specified, we update the adapter to rely on Nx's workspace analysis mechanism to read the builder that was specified (which can only happen post-installation of the Nx dependency).


packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -10,22 +10,28 @@ import {
 
 const root = process.cwd();
 
-// Determine root of project to build
+// Determine which build runner to use
+let cmd = process.env.MONOREPO_COMMAND || DEFAULT_COMMAND;
+
+// Read environment variable (only relevant for monorepos with multiple targets)
```
Yuangwang, line 16: Could you note what target means in this case in the comment too, I think it seems to indicate if the project is a monorepo or not, but that's not entirely clear

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -10,22 +10,28 @@ import {
...
+let target = process.env.MONOREPO_PROJECT || "";
```
Yuangwang, line 17: const instead of let

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -100,11 +106,12 @@ export function populateOutputBundleOptions(outputPaths: OutputPaths): OutputBun
 export const build = (
   projectRoot = process.cwd(),
   cmd = DEFAULT_COMMAND,
+  ...argv: string[]
 ): Promise<OutputBundleOptions> =>
   new Promise((resolve, reject) => {
     // enable JSON build logs for application builder
     process.env.NG_BUILD_LOGS_JSON = "1";
-    const childProcess = spawn(cmd, ["run", "build"], {
+    const childProcess = spawn(cmd, ["build", ...argv], {
```
Yuangwang, line 114: if cmd = DEFAULT_COMMAND then this line will run ```npm build```. Is that a valid command? I think you might run into this error ```Unknown command: "build"```

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -10,22 +10,28 @@ import {
 
 const root = process.cwd();
 
-// Determine root of project to build
+// Determine which build runner to use
+let cmd = process.env.MONOREPO_COMMAND || DEFAULT_COMMAND;
```
Yuangwang, line 14: Could we fetch all of the Monorepo related env vars all at once maybe in some function and have it all documented there? I'm having a bit of a hard time following all of the different monorepo related env vars.

Also it might be nice to have the Monorepo related code extracted out to a new file in general with their own set of unit tests too especially if this may get more complex as time goes on. Doesn't have to be in the pr, but might be good to add a TODO

.gitignore
```diff
@@ -1,6 +1,8 @@
 /packages/**/*/dist/
 /examples/
 /dist
+/packages/@apphosting/publish-dev/.local
```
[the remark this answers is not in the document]
blidd-google: I added this in case anyone wanted to use the local adapter publishing workflow, which will generate this directory to store downloaded npm modules.

packages/@apphosting/adapter-angular/src/bin/build.ts
blidd-google: Yep - I've also renamed the env var to "project" for more clarity

packages/@apphosting/adapter-angular/src/utils.ts
blidd-google: Ah yeah good catch - I thought `npm build` was still a valid alias for `npm run build` but not anymore. Will adjust accordingly

packages/@apphosting/adapter-angular/src/bin/build.ts
blidd-google: Yes this is absolutely something that should be done. Unfortunately there's a non-trivial amount of work to refactor this code into another module that can be imported by all adapters. In the interest of getting these fixes out in time for I/O I haven't addressed it in this PR, but it's at the top of my TODO list for sure

tonyjhuang (commented):
Thanks for the fix Brian! I do think this change adds some additional tech debt to na already complex part of the flow, but given the time constraints I don't think it's feasible to address it now.

My one ask would be to add documentation where needed to detail any tradeoffs in design and potential cleanups so that this can be easily refactored in the future. Feel free to submit whenever Wei stamps LGTM

Yuangwang (approved):

blidd-google (commented):
Yep, a refactor to clean up some of the tech debt will be a high priority item for me post-I/O. Our bug bashing revealed some of the improvements we could make to the design and once our changes have become stable it's 100% worth a revisit.

PR #180, opened 2024-05-09
Run Nx + Next.js builds from root context

blidd-google:
Companion PR to https://github.com/FirebaseExtended/firebase-framework-tools/pull/177 for Nx + Next.js builds. One minor difference is that the `FIREBASE_APP_DIRECTORY` environment variable will still need to be specified, because the server.js file will be located at `standalone/<FIREBASE_APP_DIRECTORY>/server.js` in the .next standalone output directory and we need to specify that path in the bundle.yaml.


packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -16,13 +16,19 @@ if (process.env.FIREBASE_APP_DIRECTORY) {
   projectRoot = projectRoot.concat("/", process.env.FIREBASE_APP_DIRECTORY);
 }
 
-// Determine which build runner to use
-let cmd = DEFAULT_COMMAND;
-if (process.env.MONOREPO_COMMAND) {
-  cmd = process.env.MONOREPO_COMMAND;
+// Determine which command to run the build
```
Yuangwang, line 19: Could you move this to line 31 right above where its first used?

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -16,13 +16,19 @@ if (process.env.FIREBASE_APP_DIRECTORY) {
...
+const cmd = process.env.MONOREPO_COMMAND || DEFAULT_COMMAND;
+
+// Parse args to pass to the build command
+let cmdArgs: string[] = [];
+if (process.env.MONOREPO_BUILD_ARGS) {
```
Yuangwang, line 24: Where is MONOREPO build args set? I dont see it anywhere in the buildpacks code either

blidd-google: MONOREPO_BUILD_ARGS is a new environment variable added in cl/631875763

Yuangwang: Maybe I'm missing something, but I can't seem to find MONOREPO_BUILD_ARGS in that cl. Could you help point me to it?

blidd-google: Whoops sorry, just uploaded the changes - should be in line 95 of `firebasenx/main.go` now

Yuangwang (approved):

PR #187, opened 2024-05-10
Fix pnpm build for angular

Yuangwang:
Check one layer up in require.resolve

### DIFF
--- a/packages/@apphosting/adapter-angular/src/utils.ts
+++ b/packages/@apphosting/adapter-angular/src/utils.ts
@@ -32,15 +32,15 @@ export const REQUIRED_BUILDER = "@angular-devkit/build-angular:application";
  */
 export async function checkStandaloneBuildConditions(cwd: string): Promise<void> {
   // dynamically load Angular so this can be used in an NPX context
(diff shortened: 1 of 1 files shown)


sjjj986 (approved):

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -32,14 +32,17 @@ export const REQUIRED_BUILDER = "@angular-devkit/build-angular:application";
  */
 export async function checkStandaloneBuildConditions(cwd: string): Promise<void> {
   // dynamically load Angular so this can be used in an NPX context
+  const angularCorePath = require.resolve("@angular/core", { paths: [process.cwd()] });
```
jamesdaniels, line 35: Can you use the cwd var?

jamesdaniels (approved):
See comment about reusing the existing constant otherwise LGTM

packages/@apphosting/adapter-angular/src/utils.ts
Yuangwang: Done

PR #193, opened 2024-05-16
Repro lru-break

jamesdaniels:


packages/@apphosting/adapter-angular/rollup.config.js
```diff
@@ -0,0 +1,12 @@
+import resolve from "@rollup/plugin-node-resolve";
+import json from "@rollup/plugin-json";
+import commonjs from "rollup-plugin-commonjs";
+
+export default {
+  input: "./dist/simple-server/server.js",
+  output: {
+    file: "./dist/simple-server/bundled_server.mjs",
```
Yuangwang, line 8: Given this output file path has changed, does https://github.com/FirebaseExtended/firebase-framework-tools/blob/main/packages/%40apphosting/adapter-angular/src/utils.ts#L218 still work?

jamesdaniels: It should ultimately end up in the same spot in the dist directory after this change. Rollup was run manually ( before tsc) it was then copied over by tsc (`src/simple-server/*`), now that I've converted simple-server/server to typescript (and changed to `src/simple-server/*.ts`) tsc will produce the .js file in the dist directory and we can do the rollup post-tsc by reaching into that directory.

Yuangwang: That makes sense to me in theory, but do you mind also giving it a run on an angular-csr app also? (Maybe you've already done this and it works fine) We don't really have unit test coverage around here so those are unlikely to find the issue and I want to be a bit more careful now that we've launched

jamesdaniels: SG. I'll give it a manual verification

jamesdaniels: Added functional test coverage over the adapter

packages/@apphosting/adapter-angular/src/simple-server/server.ts
```diff
@@ -1,12 +1,24 @@
...
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 const oneYear = 31_536_000_000;
 
+const prerenderedRoutesPath = path.join(__dirname, "..", "prerendered-routes.json");
+const prerenderedRoutes = existsSync(prerenderedRoutesPath)
+  ? JSON.parse(readFileSync(prerenderedRoutesPath, "utf8"))
+  : {};
+
+const staticWithoutCacheControl = express.static(path.join(__dirname, "..", "browser"));
```
jamesdaniels, line 16: No cache control is a little agressive, maybe set to 5 minutes or something more reasonable

packages/@apphosting/adapter-angular/src/simple-server/server.ts
```diff
@@ -0,0 +1,30 @@
...
+
+prerenderedRoutes.routes?.forEach((route: string) => {
+  app.get(route, staticWithoutCacheControl);
+});
+
+app.use(express.static(path.join(__dirname, "..", "browser"), { maxAge: oneYear }));
+app.get("*", (request, response) => {
+  response.sendFile(path.join(__dirname, "..", "browser", "index.html"));
+});
```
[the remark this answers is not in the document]
jamesdaniels: Since this is a static, single file, not dynamic we can probably ignore this if express caches; else we should read into memory.

PR #195, opened 2024-05-22
Astro support POC

Yuangwang:
Migrate the logic from web framework into an adapter that is published to npm.

Goes together with this pr https://github.com/firebase/firebase-tools/pull/7210. This current pr should be merged in before https://github.com/firebase/firebase-tools/pull/7210 is merged


packages/@apphosting/adapter-astro/package.json
```diff
@@ -0,0 +1,64 @@
...
+      "dist"
+    ],
+    "license": "Apache-2.0",
+    "dependencies": {
+      "fs-extra": "^11.1.1",
+      "yaml": "^2.3.4",
+      "semver": "*"
+    },
+    "peerDependencies": {
+      "next": "~14.0.0"
```
leoortizz, line 48: this should be astro instead?

packages/@apphosting/adapter-astro/src/interface.ts
```diff
@@ -0,0 +1,14 @@
+export interface AstroConfig {
```
leoortizz, line 1: wonder if we can get this type directly from Astro?

Yuangwang: This Astro config is a modified version of the one supplied by Astro, we want the modified one so I think writing our own makes sense here.

packages/@apphosting/adapter-astro/package.json
Yuangwang: Good catch, changed this to astro

PR #198, opened 2024-06-07
Nuxt adapter [WIP]

leoortizz:


packages/@apphosting/adapter-nuxt/src/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+async function getWantsBackend(cwd: string): Promise<boolean> {
+  const { ssr: wantsBackend } = await getConfig(cwd);
+
+  return wantsBackend;
+}
+
+export async function build(cwd: string, cmd = DEFAULT_COMMAND): Promise<void> {
+  const command = (await getWantsBackend(cwd)) ? "build" : "generate";
+
+  const build = spawnSync(cmd, ["run", command], {
```
Yuangwang, line 30: having cmd and command both be variables here is a bit confusing, could you rename one of them?

packages/@apphosting/adapter-nuxt/src/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+export async function generateOutputDirectory(
+  cwd: string,
+  outputBundleOptions: OutputBundleOptions,
+): Promise<void> {
+  const outDir = join(cwd, ".output");
+
+  await fsExtra.copy(outDir, outputBundleOptions.outputDirectory, {
+    overwrite: true,
+  });
+  // TODO: review, is this needed? Currently having errors with this
```
Yuangwang, line 55: We can leave this out for now if it is not needed for web frameworks cases

packages/@apphosting/adapter-nuxt/src/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+  // await fsExtra.copy(
+  //   join(cwd, "node_modules"),
+  //   join(outputBundleOptions.outputDirectory, "node_modules"),
+  //   { overwrite: true },
+  // );
+
+  await Promise.all([generateBundleYaml(outputBundleOptions, cwd)]);
```
Yuangwang, line 62: Do we need promise.all if this is just a single promise?

packages/@apphosting/adapter-nuxt/src/interfaces.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Options to help generate output directory
+ */
+export interface OutputBundleOptions {
+  bundleYamlPath: string;
+  outputDirectory: string;
+  wantsBackend: boolean;
+  clientDirectory: string;
+  serverDirectory: string;
+
+  // TODO: review, could not find yet
```
Yuangwang, line 11: We might want it for later to support App Hosting use cases but for now this is not needed. For future reference though, does nuxt have a server we can call `node ./server-file` on?

packages/@apphosting/adapter-nuxt/src/utils.ts
leoortizz: definitely no, moved the output copy here

packages/@apphosting/adapter-nuxt/src/utils.ts
leoortizz: indeed, renamed some things here to avoid confusion

packages/@apphosting/adapter-nuxt/src/interfaces.ts
leoortizz: added `server/index.mjs` per the docs: https://nuxt.com/docs/getting-started/deployment#entry-point, although it didn't work in the test app I have for some reason, might be an issue with the app I'm using

Yuangwang (approved):

cainenielsen:
Where are we at with this update? Really excited for native Nuxt support.

PR #199, opened 2024-06-12
fix application builder link

sjjj986:
This application builder link directs users to angular.dev to migrate existing Angular applications to use the new Applciation Builder system, which App Hosting requires. The old link deprecated, now it's at https://angular.dev/tools/cli/build-system-migration#for-existing-applications


Yuangwang (approved):
LGTM, could you add a brief description too? In case we ever look at this pr later and wonder what its doing at a glance

PR #204, opened 2024-06-17
Refactor adapter build configuration logic into @apphosting/common module

blidd-google:
Burning down on some tech debt here from the pre-I/O push. Refactors the logic configuring build options into an npm module that can be published separately and shared between the adapters.


packages/@apphosting/common/src/index.ts
```diff
@@ -0,0 +1,25 @@
+// Options to configure the build of a framework application
+export interface BuildOptions {
+  // command to run build script (e.g. "npm", "nx", etc.)
```
tonyjhuang, line 3: This may be too early to discuss but does this support go frameworks?

taeold (approved):

packages/@apphosting/common/src/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  monorepoProject?: string;
+}
+
+export const DEFAULT_COMMAND = "npm";
+
+// Get a set of default options, derived from the environment variable API
+// passed down to the adapter from the buildpacks environment.
+export function getBuildOptions(): BuildOptions {
+  return {
+    buildCommand: process.env.MONOREPO_COMMAND || DEFAULT_COMMAND,
```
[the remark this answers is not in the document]
blidd-google: Publicly? I have a doc here: go/apphosting-monorepo-x-framework. It's a bit outdated, but it should be the canonical reference for the Monorepo Buildpack X Framework Adapter API.

packages/@apphosting/common/src/index.ts
blidd-google: At first glance I don't why not - the BuildOptions interface is essentially just allowing something (whether that's the monorepo buildpack or a go buildpack) to configure how the build is run.

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -4,35 +4,22 @@ import {
...
-
-// Determine which command to run the build
-const cmd = process.env.MONOREPO_COMMAND || DEFAULT_COMMAND;
+const opts = getBuildOptions();
 
 // Run build command from the subdirectory if specified.
 // N.B. We run the build command from the root for monorepo builds, so that the build process can
 // locate necessary files outside the project directory.
-build(process.env.MONOREPO_CMD ? root : projectRoot, cmd, ...cmdArgs);
+build(root, opts);
```
Yuangwang, line 18: Would it be possible to simply this contract more on the adapter side? (This would probably complicate the getBuildOptions function).

I'm envisioning something like a new function buildMonorepoIfNeeded, that will handle 100% of monorepo related things, including the nx/turbo/... build. That way the adapter authors would need to know less about monorepos. This would help greatly in the case of community adapters where I'd be worried people might find monorepo support confusing.

I'm imagining something like this:

buildMonorepoIfNeeded(...){
  // pull in the env vars the same way getBuildOptions does
  // run the monorepo build if needed
  return didMonorepoBuildRun, projectDirectory 
}
then later to use it:
didMonorepoBuildRun, projectDirectory = buildMonorepoIfNeeded(...)

if !didMonorepoBuildRun{
  build(root, opts)
}
then the rest is the same.

I like this way better since its more contained within a few lines and I think it might be easier for our docs to provide an example of how to use this

Do you think this would be possible and generalizable to other monorepo frameworks too? Also open to other ideas too if you think this can be simplified another way.

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -1,38 +1,25 @@
...
 
-let projectRoot = root;
-if (process.env.FIREBASE_APP_DIRECTORY) {
-  projectRoot = join(root, process.env.FIREBASE_APP_DIRECTORY);
-}
+// Set standalone mode
+process.env.NEXT_PRIVATE_STANDALONE = "true";
+// Opt-out sending telemetry to Vercel
+process.env.NEXT_TELEMETRY_DISABLED = "1";
+await runBuild();
```
tonyjhuang, line 18: Do we want to support adapter authors who want to run a customized build command? Do we know that env vars are always sufficient to configure the build appropriately?

tonyjhuang: Could you put together a small discovery doc of how other frameworks build commands work and whether this model would support them?

That would really help me understand the tradeoffs of this design

packages/@apphosting/adapter-angular/src/bin/build.ts
```diff
@@ -1,48 +1,24 @@
...
 // Check build conditions, which vary depending on your project structure (standalone or monorepo)
-if (project) {
-  checkMonorepoBuildConditions(cmd, project);
-} else {
-  await checkStandaloneBuildConditions(projectRoot);
-}
+await checkBuildConditions(opts);
 
-const outputBundleOptions = await build(projectRoot, cmd, ...cmdArgs);
+process.env.NG_BUILD_LOGS_JSON = "1";
```
Yuangwang, line 16: Could you leave a comment for why this is needed

packages/@apphosting/adapter-nextjs/src/bin/build.ts
Yuangwang: +1 would love to see a doc that outlines this runBuild/ env var strategy so we can use it for our community framework strategy too.

That said I think it makes sense to publish this pr and continue the discussion elsewhere for now to unblock nx support.

Yuangwang (approved):
lgtm overall, only one small nit

PR #205, opened 2024-06-18
fix SvelteKit server responses

leoortizz:
Fix SvelteKit server responses. Huge thanks to @m3tasploit for the initiative. This PR differs from #159 by always responding with a Buffer.

Created test deployments here:
- With this fix: https://fb-tools-dev--svelkite-buffer-test-9b25lhcv.web.app/
- With the current production server: https://fb-tools-dev--svelkite-buffer-test-prod-6sdoxhqh.web.app/

Click in the `request` buttons to test requests. Note that images are broken with the current production server.

### DIFF
--- a/packages/firebase-frameworks/src/sveltekit/index.ts
+++ b/packages/firebase-frameworks/src/sveltekit/index.ts
@@ -26,21 +26,11 @@ export const handle = async (req: Request, res: Response) => {
     return res.writeHead(404, "Not Found").end();
   }
 
-  let body;
-  const contentType = rendered.headers.get("Content-Type");
-  if (
-    contentType.startsWith("text/") ||
-    contentType.startsWith("application/json") ||
-    contentType.startsWith("application/xml") ||
-    contentType.startsWith("application/javascript") ||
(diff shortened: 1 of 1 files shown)


jamesdaniels (approved):

packages/firebase-frameworks/src/sveltekit/index.ts
```diff
@@ -26,8 +26,11 @@ export const handle = async (req: Request, res: Response) => {
     return res.writeHead(404, "Not Found").end();
   }
 
-  const body = await rendered.text();
-  return res.writeHead(rendered.status, Object.fromEntries(rendered.headers)).end(body);
+  const body = (await rendered.arrayBuffer()) as ArrayBuffer;
```
jamesdaniels, line 29: nit, can we simply to `const body: ArrayBuffer = await rendered.arrayBuffer();`

leoortizz: done, although the types coming from `rendered` are all `any`

PR #211, opened 2024-07-02
Add env var section

sjjj986:


packages/@apphosting/adapter-angular/src/interface.ts
```diff
@@ -12,6 +12,12 @@ export interface OutputBundleOptions {
   needsServerGenerated: boolean;
 }
 
+export interface EnvironmentVariable {
```
tonyjhuang, line 15: comments for exported interfaces please

packages/@apphosting/adapter-angular/src/interface.ts
```diff
@@ -12,6 +12,12 @@ export interface OutputBundleOptions {
...
+  variable: string;
+  value: string;
+  avalability: string[];
```
tonyjhuang, line 18: typo: availability

tonyjhuang, line 18: string or enum?

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -194,20 +195,34 @@ export async function generateOutputDirectory(
   await generateBundleYaml(outputBundleOptions, cwd);
 }
 
+const runtimeEnvVars: EnvironmentVariable[] = [];
```
tonyjhuang, line 198: Why does this need to be a file level variable?

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -194,20 +195,34 @@ export async function generateOutputDirectory(
...
+
+// add environment variable to bundle.yaml if needed for specific versions
+function addBundleYamlEnvVar(): void {
+  const ssrPortEnvVar: EnvironmentVariable = {
```
tonyjhuang, line 202: suggestion: I would inline this with a comment explaining why we need this env var

tonyjhuang: Essentially, the question you want to answer is "why are we adding this env var for this version?"

Consider how you would explain this env var to a maintainer looking at this code a year from now

packages/@apphosting/adapter-angular/src/bin/build.spec.ts
```diff
@@ -35,14 +35,13 @@ describe("build commands", () => {
     const expectedFiles = {
       ".apphosting/dist/browser/browserfile": "",
       ".apphosting/dist/server/server.mjs": "",
-      ".apphosting/bundle.yaml": `headers: []
-redirects: []
-rewrites: []
+      ".apphosting/bundle.yaml": `
```
tonyjhuang, line 38: Can we add another test that SSR_PORT is added for v17.3.2

tonyjhuang (changes requested):
Thanks for the PR! added some comments

packages/@apphosting/adapter-angular/src/interface.ts
sjjj986: I think we should stay consistent with the publisher schema, so using string here: https://source.corp.google.com/piper///depot/google3/third_party/gcp_buildpacks/pkg/firebase/apphostingschema/apphostingschema.go;l=57;rcl=652323474

tonyjhuang: I think the use cases are slightly different; the publisher uses strings because it's parsing user input which doesn't necessarily have the ability to leverage enums (since it's originally just yaml format).

Besides consistency, is there a reason to keep these as strings? I would argue that constraining these to enums may help us ensure that adapter authors don't accidentally pass us the wrong values.

See http://principles-wiki.net/principles:easy_to_use_and_hard_to_misuse#:~:text=A%20module%20is%20easy%20to,the%20compiler%20signals%20wrong%20usages.

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: good catch, added a comment "add env var to solve angular port issue, existing only for Angular v17.3.2 (b/332896115)", this bug linked should have all contexts needed

packages/@apphosting/adapter-angular/src/bin/build.spec.ts
sjjj986: added

packages/@apphosting/adapter-angular/src/interface.ts
sjjj986: Hmm that makes sense, I guess besides consistency they needs to be merged in the publisher so thought it'd be nice to follow the same interface. But enums sounds more accurate in the adapter author case, changing it to enums.

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -143,30 +149,46 @@ function extractManifestOutput(output: string): string {
...
+    runtimeEnvVars.push(ssrPortEnvVar);
+  }
+  return runtimeEnvVars;
 }
 
 // Generate bundle.yaml
 async function generateBundleYaml(
   outputBundleOptions: OutputBundleOptions,
   cwd: string,
+  angualrVersion?: string,
```
Yuangwang, line 182: nit: angular typo

packages/@apphosting/adapter-angular/src/interface.ts
```diff
@@ -12,6 +12,19 @@ export interface OutputBundleOptions {
...
+export interface EnvironmentVariable {
+  variable: string;
+  value: string;
+  availability: availability[]; // currently support RUNTIME only
+}
+
+// defines whether the environment variable is buildtime, runtime or both
+export enum availability {
+  BUILDTIME = "BUILDTIME",
+  RUNTIME = "RUNTIME",
```
Yuangwang, line 25: If we only support RUNTIME should we reflect that in availability too? So we'd have RUNTIME be the only option

sjjj986: Sounds good, I changed the env var definition to have runtime only env var rn

Yuangwang (approved):

packages/@apphosting/adapter-angular/src/interface.ts
```diff
@@ -12,6 +12,19 @@ export interface OutputBundleOptions {
...
 
+// Environment variable schema for bundle.yaml outputted by angular adapter
+export interface EnvironmentVariable {
+  variable: string;
+  value: string;
+  availability: availability.RUNTIME; // currently support RUNTIME only
+}
+
+// defines whether the environment variable is buildtime, runtime or both
+export enum availability {
```
tonyjhuang, line 23: Should this be capitalized? eg Availability

tonyjhuang (approved):

PR #213, opened 2024-07-09
Set up publishing for @apphosting/common

blidd-google:


jamesdaniels (approved):

packages/@apphosting/common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@apphosting/common",
-  "version": "1.0.0",
+  "version": "0.0.1",
```
Yuangwang, line 3: You'll probably want to change both adapters to use this new version too

blidd-google: will do in separate PRs!

PR #215, opened 2024-07-09
Update @apphosting/common dependency in adapters

blidd-google:
The only thing that's changing in the Angular adapter's package.json is the @apphosting/common dependency - there's just some weirdness going on with the formatter that's setting it back to 2 space from 4.


Yuangwang:
Could you double check why the build check if failing? This change looks like it shouldn't break things but would be good to confirm

Yuangwang (approved):

PR #227, opened 2024-07-14
Update sharp version

Snugug:
The current sharp peer dependency is out of date. If you're trying to deploy a function that would use sharp, the deploy errors because there's a version mismatch. This is because SEMVER < 1 counts minor releases (0.xx) as potentially breaking, so `0.32` won't resolve with `0.33.x` installed

### DIFF
--- a/packages/firebase-frameworks/package.json
+++ b/packages/firebase-frameworks/package.json
@@ -55,7 +55,7 @@
   "peerDependencies": {
     "firebase": "^9.9.0 || ^10.0.0",
     "firebase-admin": "^11.0.1 || ^12.0.0",
-    "sharp": "^0.32.1"
+    "sharp": "^0.33.4"
   },
   "peerDependenciesMeta": {
     "firebase": {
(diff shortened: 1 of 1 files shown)


emwp:
Opened a PR on Firebase Tools to also update sharp version there: 
https://github.com/firebase/firebase-tools/pull/7517

leoortizz (approved):

packages/firebase-frameworks/package.json
```diff
@@ -55,7 +55,7 @@
   "peerDependencies": {
     "firebase": "^9.9.0 || ^10.0.0",
     "firebase-admin": "^11.0.1 || ^12.0.0",
-    "sharp": "^0.32.1"
+    "sharp": "^0.33.4"
```
leoortizz, line 58: just realized ^0.33.4 would break the build for people using old CLI versions after seeing the E2E test failure in the CLI PR(https://github.com/firebase/firebase-tools/pull/7517#pullrequestreview-2219342955), to keep it backwards compatible we need to support both versions
```suggestion
    "sharp": "^0.32 || ^0.33"
```

leoortizz (changes requested):

leoortizz (approved):

PR #228, opened 2024-07-29
Match framework version with buildpack changes

sjjj986:
Change env var representing framework version all to "FRAMEWORK_VERSION"; and buildtime availability to "BUILD"


Yuangwang (approved):
LGTM, just one note is that you should be careful with the release ordering here. You'll want to make sure that this change goes out at the same time as cl/655660411 to make sure that the env vars stay in line

PR #235, opened 2024-08-05
Astro/Angular express fix

leoortizz:


leoortizz:
the issue was local only, not needed anymore

PR #236, opened 2024-08-05
PoC of detected an unguarded ng run()

jamesdaniels:


packages/firebase-frameworks/src/angular/index.ts
```diff
@@ -1,14 +1,29 @@
...
-import { handle as expressHandle } from "../express/index.js";
 import { basename, join, normalize, relative } from "path";
 import { createReadStream } from "fs";
 import { mediaTypes } from "@hapi/accept";
+import { pathToFileURL } from "url";
 
 const LOCALE_FORMATS = [/^ALL_[a-z]+$/, /^[a-z]+_ALL$/, /^[a-z]+(_[a-z]+)?$/];
 const NG_BROWSER_OUTPUT_PATH = process.env.__NG_BROWSER_OUTPUT_PATH__;
 
+const expressHandle = new Promise<(typeof import("../express/index.js"))["handle"]>((resolve) => {
```
jamesdaniels, line 12: With this proven, we can probably just fix this in the bootstrap.js file

9kubczas4: I tried something similar but without success, but your proposal works, I created PR with update for bootstrap script: https://github.com/firebase/firebase-tools/pull/7556.

Please take a look if you'll have time.

PR #239, opened 2024-08-06
Fix for workspaces, bring back in npm i skip

jamesdaniels:
Found the root cause of the firebase-frameworks issues, the NPM workspaces were defined incorrectly.


jamesdaniels:
Ok seems the problem with lru-cache still exists

PR #240, opened 2024-08-07
manually bump firebase-frameworks version

chalosalvador:


packages/firebase-frameworks/package-lock.json
leoortizz, line 1: since it's a monorepo, npm install should be ran in the root folder. There should be only one package-lock in the repo, so we should remove this one

leoortizz (changes requested):

leoortizz (approved):

PR #244, opened 2024-08-08
Add output bundle spec to common module

Yuangwang:
This interface will be used in the other adapters as the output bundle

### DIFF
--- a/packages/@apphosting/common/src/index.ts
+++ b/packages/@apphosting/common/src/index.ts
@@ -50,7 +50,7 @@ interface EnvVarConfig {
 
 // Represents where environment variables are made available
 enum Availability {
-  // Runtime environment variables are available on the server when the app is run 
+  // Runtime environment variables are available on the server when the app is run
   Runtime,
 }
 
--- a/packages/@apphosting/common/src/index.ts
+++ b/packages/@apphosting/common/src/index.ts
@@ -1,7 +1,7 @@
 import { spawn } from "child_process";
 
 // Output bundle metadata specifications to be written to bundle.yaml
-export interface OutputBundleConfigs {
+export interface OutputBundleConfig {
   version: "v1";
   serverConfig: ServerConfig;
   metadata: Metadata;
@@ -9,25 +9,20 @@ export interface OutputBundleConfigs {
 
 // Fields needed to configure the App Hosting server
 interface ServerConfig {
-  // Command to start the server (e.g. ["node", "dist/index.js"]). Assume this command is run from the root dir of the workspace
-  runCommand: string[];
+  // Command to start the server (e.g. "node dist/index.js"). Assume this command is run from the root dir of the workspace
+  runCommand: string;
   // Environment variables set when the app is run
   environmentVariables?: EnvVarConfig[];
+  // See https://firebase.google.com/docs/reference/apphosting/rest/v1beta/projects.locations.backends.builds#runconfig for documentation on the next fields
   // The maximum number of concurrent requests that each server instance can receive.
-  // Defaults to 80.
   concurrency?: number;
   // The number of CPUs used in a single server instance.
-  // Defaults to 1.
-  cpu?: "fractional" | 1 | 2 | 4 | 6 | 8;
+  cpu?: number;
   // The amount of memory available for a server instance.
-  // Commonly one of: "256" | "512" | "1024" | "2048" | "4096" | "8192" | "16384"
-  // Defaults to 512MiB.
   memoryMiB?: string;
   // The limit on the minimum number of function instances that may coexist at a given time.
-  // Defaults to 0.
   minInstances?: number;
   // The limit on the maximum number of function instances that may coexist at a given time.
-  // Defaults to 100.
   maxInstances?: number;
 }
 
@@ -50,11 +45,12 @@ interface EnvVarConfig {
   // Value associated with the variable
   value: string;
   // Where the variable will be available, for now only RUNTIME is supported
-  availability: Availabilities.Runtime[];
+  availability: Availability.Runtime[];
 }
 
-// Represents where environment variables are available from
-enum Availabilities {
+// Represents where environment variables are made available
+enum Availability {
+  // Runtime environment variables are available on the server when the app is run 
   Runtime,
 }
 
--- a/packages/@apphosting/common/src/index.ts
+++ b/packages/@apphosting/common/src/index.ts
@@ -1,7 +1,7 @@
 import { spawn } from "child_process";
 
 // Output bundle metadata specifications to be written to bundle.yaml
-export interface OutputBundleConfigs{
+export interface OutputBundleConfigs {
   version: "v1";
   serverConfig: ServerConfig;
   metadata: Metadata;
@@ -35,7 +35,7 @@ interface ServerConfig {
 interface Metadata {
   // Name of the adapter (this should be the official package name)
   adapterPackageName: string;
-  // Version of the adapter 
+  // Version of the adapter
   adapterVersion?: string;
   // Name of the framework that is being supported, e.g. "angular"
   framework: string;
--- a/packages/@apphosting/common/src/index.ts
+++ b/packages/@apphosting/common/src/index.ts
@@ -19,7 +19,7 @@ interface ServerConfig {
   // The number of CPUs used in a single server instance.
   cpu?: number;
   // The amount of memory available for a server instance.
-  memoryMiB?: string;
+  memoryMiB?: number;
   // The limit on the minimum number of function instances that may coexist at a given time.
   minInstances?: number;
   // The limit on the maximum number of function instances that may coexist at a given time.
--- a/packages/@apphosting/common/src/index.ts
+++ b/packages/@apphosting/common/src/index.ts
@@ -1,5 +1,58 @@
 import { spawn } from "child_process";
 
+// Output bundle specifications to be written to bundle.yaml
+export interface OutputBundle {
+  version: "v1alpha";
+  serverConfig: ServerConfig;
+  metadata: Metadata;
+}
+
+// Fields needed to configure the App Hosting server
+interface ServerConfig {
+  // Command to start the server (e.g. ["node", "dist/index.js"]) assume this command is run from the root dir of the workspace
+  runCommand: string[];
+  // Environment variable present in the server execution environment.
+  environmentVariables?: EnvVarConfig[];
+  // The maximum number of concurrent requests that each server instance can receive.
+  // Defaults to 80.
+  concurrency?: number;
+  // The number of CPUs used in a single server instance.
+  // Defaults to 1.
+  cpu?: "fractional" | 1 | 2 | 4 | 6 | 8;
+  // The amount of memory available for a server instance.
+  // Commonly one of: "256" | "512" | "1024" | "2048" | "4096" | "8192" | "16384"
+  // Defaults to 512MiB.
+  memoryMiB?: string;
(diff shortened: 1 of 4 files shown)


packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
...
+// Additonal fields needed for adapter identification
+interface Metadata {
+  // Name of the adapter (this should be the npm package name)
+  adapterNpmPackageName: string;
+  // Version of the adapter
+  adapterVersion?: string;
+  // Name of the framework that is being supported
+  framework: string;
+  // Version of the framework that is being supported
+  frameworkVersion?: string;
```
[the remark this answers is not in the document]
Yuangwang: Yes that's exactly right

packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
 import { spawn } from "child_process";
 
+// Output bundle specifications to be written to bundle.yaml
+export interface OutputBundle {
```
tonyjhuang, line 4: So one quick thought: I thought OutputBundle was the name for the entirety of the framework build process which includes:
* bundle.yaml
* static assets
* source build artifacts

Is it accurate to name this outputbundle when we're actually defining the metadata for the output bundle?

packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
...
+  version: "v1alpha";
```
tonyjhuang, line 5: prefer to not use alpha/beta versioning. it's being phased out at google and it's added overhead to our own internal api release processes

packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
...
+  serverConfig: ServerConfig;
+  metadata: Metadata;
+}
+
+// Fields needed to configure the App Hosting server
+interface ServerConfig {
+  // Command to start the server (e.g. ["node", "dist/index.js"]) assume this command is run from the root dir of the workspace
```
tonyjhuang, line 12: punctation nit: "(eg ...). Assume"

packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
...
+  runCommand: string[];
```
tonyjhuang, line 13: ooc why string[] and not string?

packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
...
+  // Environment variable present in the server execution environment.
```
tonyjhuang, line 14: Is this implying Cloud Run (eg runtime)? Do we need to change this comment if we ever want to support buildtime?

packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
...
+  environmentVariables?: EnvVarConfig[];
+  // The maximum number of concurrent requests that each server instance can receive.
+  // Defaults to 80.
```
tonyjhuang, line 17: Do we want to drop these defaults now that we know GCP can adjust default values per user?

packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
...
+  concurrency?: number;
+  // The number of CPUs used in a single server instance.
```
tonyjhuang, line 19: Are we intentionally diverging from our server docs?  https://cloud.google.com/run/docs/reference/rest/v2/projects.locations.services#resource:-service

packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
...
+  memoryMiB?: string;
+  // The limit on the minimum number of function instances that may coexist at a given time.
+  // Defaults to 0.
+  minInstances?: number;
+  // The limit on the maximum number of function instances that may coexist at a given time.
+  // Defaults to 100.
+  maxInstances?: number;
+}
+
+// Additonal fields needed for adapter identification
```
tonyjhuang, line 34: "adapter identification" doesn't feel comprehensive here

packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
...
+interface Metadata {
+  // Name of the adapter (this should be the npm package name)
+  adapterNpmPackageName: string;
+  // Version of the adapter
```
tonyjhuang, line 38: Suggestion to include example strings in these docs, eg: 

"Name of the framework that is being supported, e.g. "angular""
"Version of the framework that is being supported, e.g. "18.0.1""

packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
...
+}
+
+// Configs necessary to define environment variables
+interface EnvVarConfig {
+  // Name of the variable
+  variable: string;
+  // Value associated with the variable
+  value: string;
+  // Where the variable will be available, for now this will always be RUNTIME
+  availability: "RUNTIME"[];
```
tonyjhuang, line 53: string or enum?

packages/@apphosting/common/src/index.ts
```diff
@@ -1,5 +1,58 @@
...
+  adapterNpmPackageName: string;
+  // Version of the adapter
+  adapterVersion?: string;
+  // Name of the framework that is being supported
+  framework: string;
+  // Version of the framework that is being supported
+  frameworkVersion?: string;
+}
+
+// Configs necessary to define environment variables
```
tonyjhuang, line 46: Consider: "Represents a single environment variable."

packages/@apphosting/common/src/index.ts
tonyjhuang: Hmm okay I personally would recommend we prioritize whatever makes for a better DX for adapter authors here. Splitting by " " isn't a ton of overhead on the buildpacks side, but adapter authors will have to pay the cost of consuming this API every time they add a new framework

packages/@apphosting/common/src/index.ts
tonyjhuang: What if we link to the product docs instead so we have a single source of truth?

packages/@apphosting/common/src/index.ts
```diff
@@ -31,26 +31,31 @@ interface ServerConfig {
...
-// Additonal fields needed for adapter identification
+// Additonal fields needed for identifying the framework and adapter being used
 interface Metadata {
-  // Name of the adapter (this should be the npm package name)
-  adapterNpmPackageName: string;
-  // Version of the adapter
+  // Name of the adapter (this should be the official package name) e.g. "@apphosting/adapter-nextjs"
+  adapterPackageName: string;
+  // Version of the adapter, e.g. "18.0.1"
   adapterVersion?: string;
```
tonyjhuang, line 39: why optional?

packages/@apphosting/common/src/index.ts
```diff
@@ -31,26 +31,31 @@ interface ServerConfig {
...
   // Value associated with the variable
   value: string;
-  // Where the variable will be available, for now this will always be RUNTIME
-  availability: "RUNTIME"[];
+  // Where the variable will be available, for now only RUNTIME is supported
+  availability: Availabilities.Runtime[];
+}
+
+// Represents where environment variables are available from
+enum Availabilities {
```
tonyjhuang, line 57: enum Availability

packages/@apphosting/common/src/index.ts
```diff
@@ -31,26 +31,31 @@ interface ServerConfig {
...
   variable: string;
   // Value associated with the variable
   value: string;
-  // Where the variable will be available, for now this will always be RUNTIME
-  availability: "RUNTIME"[];
+  // Where the variable will be available, for now only RUNTIME is supported
+  availability: Availabilities.Runtime[];
+}
+
+// Represents where environment variables are available from
```
tonyjhuang, line 56: comment nit: this is a bit confusing, how about "Represents where environment variables are made available."

packages/@apphosting/common/src/index.ts
```diff
@@ -31,26 +31,31 @@ interface ServerConfig {
...
+enum Availabilities {
+  Runtime,
```
tonyjhuang, line 58: We should include a doc string here

packages/@apphosting/common/src/index.ts
```diff
@@ -1,17 +1,17 @@
 import { spawn } from "child_process";
 
-// Output bundle specifications to be written to bundle.yaml
-export interface OutputBundle {
-  version: "v1alpha";
+// Output bundle metadata specifications to be written to bundle.yaml
+export interface OutputBundleConfigs {
```
tonyjhuang, line 4: nit: OutputBundleConfig

tonyjhuang (changes requested):

packages/@apphosting/common/src/index.ts
Yuangwang: Sure that makes sense, changed this to string

packages/@apphosting/common/src/index.ts
Yuangwang: Removed the defaults in favor of linking the docs.

packages/@apphosting/common/src/index.ts
Yuangwang: Oh good call, yeah this shouldn't diverge from those. Changed CPU to number

packages/@apphosting/common/src/index.ts
Yuangwang: Originally this field was added so authors had the option to tell us what version if they wanted to (technically this is something we can self discover on the buildpacks side too) but I think it's pretty straightforward for adapter authors to provide this to us as well (done here as an example https://github.com/FirebaseExtended/firebase-framework-tools/pull/248/files#diff-c4645e334899a380f959aecbba2ce11a7ffcd895b12daed80af12d265eb9d437R120-R125) 

I've gone ahead and made it required so we don't need to do it in the buildpacks anymore

packages/@apphosting/common/src/index.ts
Yuangwang: Done

packages/@apphosting/common/src/index.ts
Yuangwang: Done

packages/@apphosting/common/src/index.ts
Yuangwang: Done

packages/@apphosting/common/src/index.ts
Yuangwang: Done

packages/@apphosting/common/src/index.ts
```diff
@@ -1,56 +1,57 @@
...
-  // Defaults to 80.
   concurrency?: number;
   // The number of CPUs used in a single server instance.
-  // Defaults to 1.
-  cpu?: "fractional" | 1 | 2 | 4 | 6 | 8;
+  cpu?: number;
   // The amount of memory available for a server instance.
-  // Commonly one of: "256" | "512" | "1024" | "2048" | "4096" | "8192" | "16384"
-  // Defaults to 512MiB.
   memoryMiB?: string;
```
tonyjhuang, line 22: why string?

tonyjhuang (approved):

packages/@apphosting/common/src/index.ts
Yuangwang: Good catch, we probably want number here too

PR #246, opened 2024-08-21
Add metadata to angular bundle.yaml

sjjj986:
Add a framework metadata section to Angular adapter's bundle.yaml:
 - adapter name
 - adapter version
 - framework name
 - framework version

### DIFF
--- a/packages/@apphosting/adapter-angular/src/bin/build.spec.ts
+++ b/packages/@apphosting/adapter-angular/src/bin/build.spec.ts
@@ -46,7 +46,7 @@ staticAssets:
   - .apphosting/dist/browser
 env: []
 metadata:
-  adapterNpmPackageName: "@apphosting/adapter-angular"
+  adapterNpmPackageName: \\@apphosting/adapter-angular
   adapterVersion: ${packageVersion}
   framework: angular
   frameworkVersion: 17.3.8
@@ -79,7 +79,7 @@ env:
     value: "8080"
     availability: RUNTIME
 metadata:
-  adapterNpmPackageName: "@apphosting/adapter-angular"
+  adapterNpmPackageName: \\@apphosting/adapter-angular
   adapterVersion: ${packageVersion}
   framework: angular
   frameworkVersion: 17.3.2
--- a/packages/@apphosting/adapter-angular/src/bin/build.spec.ts
+++ b/packages/@apphosting/adapter-angular/src/bin/build.spec.ts
@@ -9,6 +9,7 @@ import { OutputBundleOptions } from "../interface.js";
 describe("build commands", () => {
   let tmpDir: string;
   let outputBundleOptions: OutputBundleOptions;
+  let defaultAngularVersion: string;
   beforeEach(() => {
     tmpDir = generateTmpDir();
     outputBundleOptions = {
@@ -20,6 +21,7 @@ describe("build commands", () => {
       serverFilePath: resolve(tmpDir, ".apphosting", "dist", "server", "server.mjs"),
       needsServerGenerated: false,
     };
+    defaultAngularVersion = "17.3.8";
   });
 
   it("expects all output bundle files to be generated", async () => {
@@ -28,9 +30,9 @@ describe("build commands", () => {
       "dist/test/browser/browserfile": "",
       "dist/test/server/server.mjs": "",
     };
-    const packageVersion = createMetadata("17.3.8").adapterVersion;
+    const packageVersion = createMetadata(defaultAngularVersion).adapterVersion;
     generateTestFiles(tmpDir, files);
-    await generateOutputDirectory(tmpDir, outputBundleOptions, "17.3.8");
+    await generateOutputDirectory(tmpDir, outputBundleOptions, defaultAngularVersion);
     await validateOutputDirectory(outputBundleOptions);
 
     const expectedFiles = {
@@ -44,7 +46,7 @@ staticAssets:
   - .apphosting/dist/browser
 env: []
 metadata:
-  adapterNpmPackageName: \\@apphosting/adapter-angular
+  adapterNpmPackageName: "@apphosting/adapter-angular"
   adapterVersion: ${packageVersion}
   framework: angular
   frameworkVersion: 17.3.8
@@ -77,7 +79,7 @@ env:
     value: "8080"
     availability: RUNTIME
 metadata:
-  adapterNpmPackageName: \\@apphosting/adapter-angular
+  adapterNpmPackageName: "@apphosting/adapter-angular"
   adapterVersion: ${packageVersion}
   framework: angular
   frameworkVersion: 17.3.2
@@ -93,7 +95,7 @@ metadata:
       "dist/test/server/notserver.mjs": "",
     };
     generateTestFiles(tmpDir, files);
-    await generateOutputDirectory(tmpDir, outputBundleOptions);
+    await generateOutputDirectory(tmpDir, outputBundleOptions, defaultAngularVersion);
     assert.rejects(async () => await validateOutputDirectory(outputBundleOptions));
   });
 
--- a/packages/@apphosting/adapter-angular/src/bin/build.spec.ts
+++ b/packages/@apphosting/adapter-angular/src/bin/build.spec.ts
@@ -23,13 +23,12 @@ describe("build commands", () => {
   });
 
   it("expects all output bundle files to be generated", async () => {
-    const { generateOutputDirectory, validateOutputDirectory, populateMetadata } =
-      await importUtils;
+    const { generateOutputDirectory, validateOutputDirectory, createMetadata } = await importUtils;
     const files = {
       "dist/test/browser/browserfile": "",
       "dist/test/server/server.mjs": "",
     };
-    const packageVersion = populateMetadata("17.3.8").adapterVersion;
+    const packageVersion = createMetadata("17.3.8").adapterVersion;
     generateTestFiles(tmpDir, files);
     await generateOutputDirectory(tmpDir, outputBundleOptions, "17.3.8");
     await validateOutputDirectory(outputBundleOptions);
@@ -55,12 +54,12 @@ metadata:
   });
 
   it("expects SSR_PORT variable is added to bundle.yaml for Angular v17.3.2", async () => {
-    const { generateOutputDirectory, populateMetadata } = await importUtils;
+    const { generateOutputDirectory, createMetadata } = await importUtils;
     const files = {
       "dist/test/browser/browserfile": "",
       "dist/test/server/server.mjs": "",
     };
-    const packageVersion = populateMetadata("17.3.2").adapterVersion;
+    const packageVersion = createMetadata("17.3.2").adapterVersion;
     generateTestFiles(tmpDir, files);
     await generateOutputDirectory(tmpDir, outputBundleOptions, "17.3.2");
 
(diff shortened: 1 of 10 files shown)


packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -142,6 +143,16 @@ function extractManifestOutput(output: string): string {
   return stripAnsi(output.substring(start, end + 1));
 }
 
+export function populateMetadata(angularVersion?: string): Metadata {
```
tonyjhuang, line 146: tsdocs for exported functions please

go/tsjs-style#document-all-top-level-exports-of-modules

Think of exported functions as your file/module/class's API. Unless it's exceedingly obvious what a function does (eg a basic getter), doc strings are usually a safe bet

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -142,6 +143,16 @@ function extractManifestOutput(output: string): string {
...
+  const packageJson = JSON.parse(readFileSync(`${__dirname}/../package.json`, "utf-8"));
```
tonyjhuang, line 147: Is this guaranteed to be the case? If this file gets moved around this will break, can we add a check or assert that package.json exists

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -142,6 +143,16 @@ function extractManifestOutput(output: string): string {
...
+  return {
+    adapterNpmPackageName: packageJson.name.replace(/([@])/g, "\\$1"), // escape @ for yamlStringify
```
tonyjhuang, line 149: can we include an example comment here, eg:

"myinitialstring" -> "myfinalstring"

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -142,6 +143,16 @@ function extractManifestOutput(output: string): string {
...
+    adapterVersion: packageJson.version,
+    framework: "angular",
+    frameworkVersion: angularVersion ? angularVersion : "",
```
tonyjhuang, line 152: Why is angularVersion nullable? When would it not be passed in? Should we be reading this from package.json or package-lock.json?

packages/@apphosting/adapter-angular/src/utils.ts
tonyjhuang, line 146: optional nit: populateMetadata sounds like it should take another metadata object as a param and assign values to fields. Can we rename this createMetadata or produceMetadata?

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -182,13 +193,15 @@ async function generateBundleYaml(
   angularVersion?: string,
 ): Promise<void> {
   const runtimeEnvVars = addBundleYamlEnvVar(angularVersion);
+  const metadata = populateMetadata(angularVersion);
```
tonyjhuang, line 196: ca we inline these two variables?

tonyjhuang (changes requested):

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: You mean like `const runtimeEnvVars = addBundleYamlEnvVar(angularVersion), metadata = createMetadata(angularVersion);`?  I think this causes a lint error, have not discovered other ways to inline these two variables.

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: `angularVersion` is the user's angular version passed from framework buildpack with `process.env.FRAMWORK_VERSION`, so is nullable. But since this should always exists, should we be asserting not null instead of defining metadata as having a nullable frameworkVersion?

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: Added

packages/@apphosting/adapter-angular/src/utils.ts
tonyjhuang: Yeah that sounds reasonable to me!

packages/@apphosting/adapter-angular/src/bin/build.spec.ts
```diff
@@ -69,6 +76,11 @@ env:
   - variable: SSR_PORT
     value: "8080"
     availability: RUNTIME
+metadata:
+  adapterNpmPackageName: \\@apphosting/adapter-angular
+  adapterVersion: ${packageVersion}
+  framework: angular
+  frameworkVersion: 17.3.2
```
tonyjhuang, line 83: I know this test was already here but can we reconfigure this testcase to ONLY test the SSR_PORT variable? As it stands, every time any change is made to the bundle.yaml output, this test needs to change which means it's testing things that it should not be

go/unit-testing-practices#behavior-testing-examples

tonyjhuang (changes requested):

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -143,10 +143,19 @@ function extractManifestOutput(output: string): string {
...
+ */
+export function createMetadata(angularVersion?: string): Metadata {
+  const packageJsonPath = `${__dirname}/../package.json`;
+  if (!existsSync(packageJsonPath)) {
+    throw new Error(`Angular adapter package.json file does not exists at ${packageJsonPath}`);
+  }
+  const packageJson = JSON.parse(readFileSync(packageJsonPath, "utf-8"));
   return {
-    adapterNpmPackageName: packageJson.name.replace(/([@])/g, "\\$1"), // escape @ for yamlStringify
+    // escape @ for yamlStringify, e.g. "@apphosting/adapter-angular" -> "\@apphosting/adapter-angular"
```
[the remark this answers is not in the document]
sjjj986: yaml.Stringify will add double quotes on adapter name because of the @:
```
metadata:
  adapterNpmPackageName: "@apphosting/adapter-angular"
  adapterVersion: 17.2.7
```
Now I rethink about this, even tho it looks slightly different from other fields, the publisher will handle this string still fine. Removing the escape here

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: Ahh got it thanks!

packages/@apphosting/adapter-angular/src/utils.ts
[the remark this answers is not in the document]
Yuangwang: imo I think we should try to keep the fields consistent and still escape the @. Especially once we open this up to more community adapters it may make analysis more complicated if we have some fields with extra quotes and some without.

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -142,6 +143,25 @@
...
+ */
+export function createMetadata(angularVersion?: string): Metadata {
+  const packageJsonPath = `${__dirname}/../package.json`;
+  if (!existsSync(packageJsonPath)) {
+    throw new Error(`Angular adapter package.json file does not exists at ${packageJsonPath}`);
+  }
+  const packageJson = JSON.parse(readFileSync(packageJsonPath, "utf-8"));
+  return {
+    // escape @ for yamlStringify, e.g. "@apphosting/adapter-angular" -> "\@apphosting/adapter-angular"
+    adapterNpmPackageName: packageJson.name.replace(/([@])/g, "\\$1"),
```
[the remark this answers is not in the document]
sjjj986: Added backslash escape: /([@\\])/g

packages/@apphosting/adapter-angular/src/bin/build.spec.ts
sjjj986: That's a good point, added a `validateFileExistsAndContains` helper and modified the env var test to only test SSR_PORT.

packages/@apphosting/adapter-angular/src/bin/build.spec.ts
```diff
@@ -56,22 +64,13 @@ env: []
...
-    const expectedFiles = {
-      ".apphosting/dist/browser/browserfile": "",
-      ".apphosting/dist/server/server.mjs": "",
-      ".apphosting/bundle.yaml": `
-runCommand: node .apphosting/dist/server/server.mjs
-neededDirs:
-  - .apphosting
-staticAssets:
-  - .apphosting/dist/browser
+    const expectedContents = `
```
tonyjhuang, line 67: optional suggestion to inline this

packages/@apphosting/adapter-angular/src/utils.ts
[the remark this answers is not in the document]
tonyjhuang: How does the yaml deserialization work on the go side? does the deserialized value include the quotes?

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -157,14 +176,15 @@ export async function generateOutputDirectory(
   if (outputBundleOptions.needsServerGenerated) {
     await generateServer(outputBundleOptions);
   }
+  if (!angularVersion) throw new Error("Could not find the angular version of the application");
```
tonyjhuang, line 179: this seems like a guard clause to me which typically happens at the top of methods: go/error-handling-tactics-control-flow#guard-clause

this way we can short circuit all of the above logic since we know this function should fail/throw an error. Also I believe we need braces for all `if` statements: go/tsjs-style#control-flow-statements-blocks

tonyjhuang, line 179: Actually, why does `angularVersion` need to be nullable here at all?

tonyjhuang (changes requested):

packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -157,14 +176,15 @@ export async function generateOutputDirectory(
...
   await generateBundleYaml(outputBundleOptions, cwd, angularVersion);
 }
 
 // add environment variable to bundle.yaml if needed for specific versions
-function addBundleYamlEnvVar(angularVersion?: string): EnvironmentVariable[] {
+function addBundleYamlEnvVar(angularVersion: string): EnvironmentVariable[] {
```
tonyjhuang, line 184: naming nit: can we change this to getEnvVars or createEnvVars? This isn't really adding env vars to bundle.yaml yet

tonyjhuang: ah i see we have `generateBundleYaml` below, how about `generateEnvVars`

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: I think process.env.FRAMEWORK_VERSION is nullable by default, and I passed that in build.ts generateOutputDirectory(root, outputBundleOptions, process.env.FRAMEWORK_VERSION). So asserting in here, is it better to assert in build.ts?

packages/@apphosting/adapter-angular/src/utils.ts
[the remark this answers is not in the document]
Yuangwang: I just double checked it in go and the deserialized value doesn't actually include the quotes. Good catch Tony.

Sorry for the back and forth, but that means that we should be good to not escape the @ then and have the quotes as is

packages/@apphosting/adapter-angular/src/utils.ts
sjjj986: Moved the assert to build.ts, makes more sense to assert `process.env.FRAMEWORK_VERSION` right before using it

tonyjhuang (approved):
Thanks for the changes!

Yuangwang (approved):

PR #248, opened 2024-08-28
Update next.js bundle.yaml with env and metadata sections

sjjj986:
- pass user's nextjs version from framework buildpack
- add an empty env section
- add a metadata section with adapter & nextjs metrics
- refactor bundle.yaml test

### DIFF
--- a/packages/@apphosting/adapter-nextjs/src/bin/build.ts
+++ b/packages/@apphosting/adapter-nextjs/src/bin/build.ts
@@ -15,7 +15,6 @@ const opts = getBuildOptions();
 process.env.NEXT_PRIVATE_STANDALONE = "true";
 // Opt-out sending telemetry to Vercel
 process.env.NEXT_TELEMETRY_DISABLED = "1";
-// Assert framework version exists
 if (!process.env.FRAMEWORK_VERSION) {
   throw new Error("Could not find the nextjs version of the application");
 }
--- a/packages/@apphosting/adapter-nextjs/src/bin/build.ts
+++ b/packages/@apphosting/adapter-nextjs/src/bin/build.ts
@@ -15,11 +15,21 @@ const opts = getBuildOptions();
 process.env.NEXT_PRIVATE_STANDALONE = "true";
 // Opt-out sending telemetry to Vercel
 process.env.NEXT_TELEMETRY_DISABLED = "1";
+// Assert framework version exists
+if (!process.env.FRAMEWORK_VERSION) {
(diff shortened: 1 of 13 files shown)


packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -15,11 +15,21 @@ const opts = getBuildOptions();
 process.env.NEXT_PRIVATE_STANDALONE = "true";
 // Opt-out sending telemetry to Vercel
 process.env.NEXT_TELEMETRY_DISABLED = "1";
+// Assert framework version exists
```
tonyjhuang, line 18: small nit: imo this comment is unnecessary, the code is straightforward enough

packages/@apphosting/adapter-nextjs/src/interfaces.ts
```diff
@@ -101,3 +101,11 @@ export interface OutputBundleOptions {
    */
   outputStaticDirectoryPath: string;
 }
+
+// Metadata schema for bundle.yaml outputted by next.js adapter
+export interface Metadata {
+  adapterNpmPackageName: string;
```
tonyjhuang, line 107: ooc how does this change when we support non node dependencies, eg for python apps?

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -112,11 +113,30 @@ async function moveResources(appDir: string, outputBundleAppDir: string): Promis
   return;
 }
 
+/**
+ * Create metadata needed for outputting adapter and framework metrics in bundle.yaml.
+ */
+export function createMetadata(nextVersion: string): Metadata {
+  const __dirname = dirname(fileURLToPath(import.meta.url));
```
tonyjhuang, line 120: why __dirname and not dirname?

tonyjhuang (changes requested):

packages/@apphosting/adapter-nextjs/src/interfaces.ts
sjjj986: We can modify the output bundle spec naming when needed? Revisiting in discussion meeting.

sjjj986: Discussed and agreed to make this field more generic, changing to`adapterPackageName`

tonyjhuang (approved):

PR #253, opened 2024-09-11
Nextjs use spec

Yuangwang:
Use the new output bundle spec for nextjs.

Also copies assets into standalone directory instead of .apphosting


packages/@apphosting/adapter-nextjs/src/bin/build.spec.ts
```diff
@@ -13,18 +13,18 @@ describe("build commands", () => {
...
     tmpDir = generateTmpDir();
     outputBundleOptions = {
-      bundleYamlPath: path.join(tmpDir, ".apphosting/bundle.yaml"),
+      bundleYamlPath: path.join(tmpDir, ".apphosting", "bundle.yaml"),
       outputDirectoryBasePath: path.join(tmpDir, ".apphosting"),
-      outputDirectoryAppPath: path.join(tmpDir, ".apphosting"),
-      outputPublicDirectoryPath: path.join(tmpDir, ".apphosting/public"),
-      outputStaticDirectoryPath: path.join(tmpDir, ".apphosting/.next/static"),
-      serverFilePath: path.join(tmpDir, ".apphosting/server.js"),
+      outputDirectoryAppPath: path.join(tmpDir, ".next", "standalone"),
```
sjjj986, line 18: why are these output paths changed to ".next/standalone"?

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -133,44 +130,32 @@ export function createMetadata(nextVersion: string): Metadata {
...
-      neededDirs: [normalize(relative(cwd, outputBundleOptions.outputDirectoryBasePath))],
-      staticAssets: [normalize(relative(cwd, outputBundleOptions.outputPublicDirectoryPath))],
-      env: [],
-      metadata: createMetadata(nextVersion),
-    }),
-  );
+  await mkdir(opts.outputDirectoryBasePath);
+  const outputBundle: OutputBundleConfig = {
+    version: "v1",
+    serverConfig: {
```
sjjj986, line 140: How will the env var section be taken cared of here?

packages/@apphosting/adapter-nextjs/src/bin/build.spec.ts
Yuangwang: We decided during the api review for the new spec to only have bundle.yaml in .apphosting. We will no longer move all the build assets into .apphosting

packages/@apphosting/adapter-nextjs/src/utils.ts
Yuangwang: It will be done the same way things are done in the angular adapter, we'll only add it if needed. For now we don't have any env var needs and env var is optional so there is no need to add a field for it yet

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -50,63 +45,65 @@ export const isMain = (meta: ImportMeta) => {
...
+  appDir: string,
+  outputBundleAppDir: string,
+  bundleYamlPath: string,
+): Promise<void> {
   const appDirExists = await exists(appDir);
   if (!appDirExists) return;
   const pathsToMove = await readdir(appDir);
   for (const path of pathsToMove) {
-    const isOutputBundleDir = join(appDir, path) === outputBundleAppDir;
+    const isbundleYamlDir = join(appDir, path) === dirname(bundleYamlPath);
```
tonyjhuang, line 104: nit: isBundleYamlDir

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -133,44 +130,32 @@ export function createMetadata(nextVersion: string): Metadata {
 
 // generate bundle.yaml
 async function generateBundleYaml(
-  outputBundleOptions: OutputBundleOptions,
-  nextBuildDirectory: string,
+  opts: OutputBundleOptions,
```
tonyjhuang, line 133: optional nit: you can unpack this if you prefer

tonyjhuang (approved):

PR #254, opened 2024-09-12
Fix apphosting/common in build/publish

jamesdaniels:
Seeing failures [since the changes to `@apphosting/common`](https://github.com/FirebaseExtended/firebase-framework-tools/actions/runs/10822830253/job/30027361965?pr=232). It turns out that in lerna, for the dependency graph to pick up dependents / dependencies the version in the package.json dependencies needs to be in the form of `*`

* Change monorepo dependencies to `*`, that's the easy part
* Now we don't want to allow any version of `@apphosting/common` as a dependency of `@apphosting/adapter-*`, so let's pin before we publish to the registry
  * Do two passes on the scopedLernaList (fka filteredLernaList)
    1. make a new package.json—set the new version and publishConfig
    1. pin dependencies to any versions from the first pass, write to disk, publish
  * Topo sort the lerna list, so packages are published in the correct order
* We should be building dependencies but only publishing dependents, move `--include-dependencies` into `build.js`
  * As build/test use the same scopedLernaList as publish, which has `--include-dependents`, changes to `@apphosting/common` will still trigger `@apphosting/adapter-*` tests to ensure we haven't introduced breaking changes
  * This went unnoticed before as dep tracking wasn't functioning
* Readability pass

Lerna publish would do much of this all for us, if we were using that... but I don't like the loss of control we'd suffer there, seems too magic.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,8 +6,8 @@
     "": {
       "name": "root",
       "workspaces": [
-        "packages/@apphosting/*",
-        "packages/create-next-on-firebase/*",
+        "packages/@apphosting/**",
+        "packages/create-next-on-firebase",
         "packages/firebase-frameworks"
       ],
       "devDependencies": {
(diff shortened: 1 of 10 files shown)


MC-MAGA:
https://kotlinlang.org/docs/releases.html#ide-support

MC-MAGA (approved):

blidd-google (approved):
Thanks for doing this - the publish flow has been _slightly_ broken since we set up the @apphosting/common package, this will be a nice QOL improvement!

PR #255, opened 2024-09-12
Add basic apphosting.yaml in next.js and angular starter templates.

taeold:


Yuangwang (approved):

starters/angular/basic/apphosting.yaml
```diff
@@ -0,0 +1,23 @@
+# Settings for Backend (on Cloud Run).
+# See https://firebase.google.com/docs/app-hosting/configure#cloud-run.
+runConfig:
+  minInstances: 0
+  # maxInstances: 100
+  # concurrency: 100
+  # cpu: 2
```
inlined, line 7: Should these values be the defaults?

starters/angular/basic/apphosting.yaml
```diff
@@ -0,0 +1,23 @@
+# Settings for Backend (on Cloud Run).
```
inlined, line 1: Maybe we should have a header explaining the use of environment-specific apphosting.yaml overrides?

starters/nextjs/basic/apphosting.yaml
```diff
@@ -0,0 +1,23 @@
+# Settings for Backend (on Cloud Run).
+# See https://firebase.google.com/docs/app-hosting/configure#cloud-run.
+runConfig:
+  minInstances: 0
```
inlined, line 4: why does Angular comment out all values but Nex.tjs does not?

inlined: I see that you use the variable in Next.js, but then why not use it in Angular SSR?

taeold: I have no idea how to use env vars in angular. If you do, welcoming a PR!

starters/angular/basic/apphosting.yaml
taeold: I think I'd rather point the users to public doc so we don't have 2 separate places to track documentations on how to use env specific overrides.

starters/angular/basic/apphosting.yaml
taeold: changed.

starters/nextjs/basic/apphosting.yaml
inlined: process.env doesn't work?

taeold: no

PR #257, opened 2024-09-15
fix: Next.js boilerplate - link to ISR on-demand page

KurtMar:
Both links on the ISR page point to the time-based example. The other link should point to the on-demand example.

Side note: There is no link from the site root page to the ISR page, which should probably be also addressed in another pull request by someone with more insight into the project.


leoortizz (approved):
Thanks!

leoortizz:
@jamesdaniels @jhuleatt thoughts on adding the ISR link back to the boilerplate homepage?

PR #261, opened 2024-09-18
Angular use spec

Yuangwang:
Use the new output bundle spec for Angular.

Also copies assets into dist directory instead of .apphosting


packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -195,20 +186,20 @@ function generateEnvVars(angularVersion: string): EnvironmentVariable[] {
 
 // Generate bundle.yaml
 async function generateBundleYaml(
-  outputBundleOptions: OutputBundleOptions,
+  opts: OutputBundleOptions,
```
tonyjhuang, line 189: optional nit: you can unpack the relevant fields here if you want

tonyjhuang (approved):

PR #276, opened 2024-11-20
Add support for angular meta framework

Yuangwang:
This pr will add additional builders allowed and skip generating the output bundle if it already exists

### DIFF
--- a/packages/@apphosting/adapter-angular/src/utils.ts
+++ b/packages/@apphosting/adapter-angular/src/utils.ts
@@ -226,3 +226,10 @@ export const isMain = (meta: ImportMeta) => {
   if (!process.argv[1]) return false;
   return process.argv[1] === fileURLToPath(meta.url);
 };
+
+export const outputBundleExists  = (outputBundleOptions: OutputBundleOptions) => {
+  if (existsSync(dirname(outputBundleOptions.bundleYamlPath))){
+    return true;
+  }
+  return false;
+}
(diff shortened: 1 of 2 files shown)


packages/@apphosting/adapter-angular/src/utils.ts
```diff
@@ -24,7 +24,10 @@ const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 const SIMPLE_SERVER_FILE_PATH = join(__dirname, "simple-server", "bundled_server.mjs");
 
-export const REQUIRED_BUILDER = "@angular-devkit/build-angular:application";
+export const ALLOWED_BUILDERS = [
```
jamesdaniels, line 27: An alternative here is to only fail if we don't see the JSON manifest & there is not an .apphosting directory at the end of the process—rather than fail early if the builder doesn't match.

Yuangwang: I double checked what would happen if we don't fail and my browser-build app will still output the JSON manifest and the .apphosting directory, but will then fail if you try to run the app since the server doesn't actually exist so I think we still need this check

jamesdaniels: Ok, thanks for giving it a shot. Maybe I'll take a pass at loosing the builder requirement next time things are a bit more quiet.

jamesdaniels (approved):

quedicesebas:
Can you elaborate more about this? In which cases the output bundle exists already?

PR #277, opened 2024-11-22
fix angular19

Yuangwang:
Add additional support for prerendered object routes

### DIFF
--- a/packages/@apphosting/adapter-angular/src/interface.ts
+++ b/packages/@apphosting/adapter-angular/src/interface.ts
@@ -63,5 +63,4 @@ export const buildManifestSchema = z.object({
     server: z.optional(url),
     browser: url,
   }),
-  prerenderedRoutes: z.optional(z.union([z.string().array(), z.object({})])),
 });
--- a/packages/@apphosting/adapter-angular/src/interface.ts
+++ b/packages/@apphosting/adapter-angular/src/interface.ts
@@ -63,5 +63,5 @@ export const buildManifestSchema = z.object({
     server: z.optional(url),
     browser: url,
   }),
-  prerenderedRoutes: z.optional(z.string().array()),
+  prerenderedRoutes: z.optional(z.union([z.string().array(), z.object({})])),
 });
--- a/packages/@apphosting/adapter-angular/src/interface.ts
+++ b/packages/@apphosting/adapter-angular/src/interface.ts
@@ -63,4 +63,7 @@ export const buildManifestSchema = z.object({
     server: z.optional(url),
     browser: url,
   }),
+  // angular v18 has an array type and v19 has an object type
+  // We should uncomment this when we need to use prerenderedRoutes
+  // prerenderedRoutes: z.optional(z.union([z.string().array(), z.object({})])),
 });


jamesdaniels:
q, are we using that object? if so maybe we limit our validation to the keys/values we care about

Yuangwang:
Good point, we don't actually use prerendered routes at all so its better to remove it.

packages/@apphosting/adapter-angular/src/interface.ts
```diff
@@ -63,5 +63,4 @@ export const buildManifestSchema = z.object({
     server: z.optional(url),
     browser: url,
   }),
-  prerenderedRoutes: z.optional(z.string().array()),
```
taeold, line 66: Do you think we can comment this out for now and add that the type is different between v19 and v18? Might be helpful later when we actually end up using this field.

taeold (approved):

packages/@apphosting/adapter-angular/src/interface.ts
Yuangwang: Done!

taeold:
@quedicesebas Sorry for the troubles. The fix needs to be rolled out to our servers, which should take sometime (ETA: first week of December).

In the meantime, we'll try to see if we can create a workaround to get angular19 working on app hosting.

quedicesebas:
I already lost 3 days with days, now I had made a rollback of the migration, so how many time more :(

Yuangwang:
@quedicesebas and anyone else having this issue. We'll keep y'all posted on when this fix goes live. Until then please give this workaround a try: 

Step 1:
In your project run `npm install @apphosting/adapter-angular@17.2.11`, this will install the latest adapter with this fix merged in.
After running that please verify that you see `@apphosting/adapter-angular": "^17.2.11` in your package.json. Its important that the version is 17.2.11 since this fix is released with that version.

Step 2:
Add this additional line in your package.json scripts: `"apphosting:build": "FRAMEWORK_VERSION=${YOUR_ANGULAR_VERSION} apphosting-adapter-angular-build"`

Step 3: 
Commit these changes to github and kickoff a new rollout

quedicesebas:
In the step 2, shouel be `"apphosting:build": "FRAMEWORK_VERSION=19.0.0 apphosting-adapter-angular-build"` for example?

Yuangwang:
yeah thats right, give it a try and lmk if it fixes your issues

quedicesebas:
Worked!

Char2sGu:
The `@apphosting/adapter-angular` depends on some outdated version of angular devkit packages.

The below dependency overrides are required to install the given version of `@apphosting/adapter-angular`:
```json
    "@apphosting/adapter-angular": {
      "@angular-devkit/architect": "$@angular-devkit/architect",
      "@angular-devkit/core": "$@angular-devkit/core"
    }
```

A `>=` semver matcher might increase the compatibility @Yuangwang

Char2sGu:
Another thing worth notice is that after installing `@apphosting/adapter-angular@17.2.11`,
dependencies are installed with `NODE_ENV` set to `production`,
while previously it was `development`.

If there are any npm `"prepare"` scripts, e.g. `husky install`, this change might cause another build error since the binary will not be installed in production.

Char2sGu:
@Yuangwang The build failed, but not because of this particular Angular 19 issue. I had to submit the report here because there is no Issues tab for this repo.

Issue:
If the output of `NG_BUILD_LOGS_JSON` is more than 2000 characters, a line break will be inserted, resulting in JSON parsing errors.

image

In my case, the warnings are mostly because of some other issues from the build pipeline, and there is no option to disable such warnings:
```
▲ [WARNING] 28 rules skipped due to selector errors:
  ::view-transition-group(*) -> Pseudo-elements are not supported by css-select
  ::view-transition-old(*) -> Pseudo-elements are not supported by css-select
  ::view-transition-new(*) -> Pseudo-elements are not supported by css-select
  ::view-transition-group(root) -> Pseudo-elements are not supported by css-select
  ::view-transition-old(header) -> Pseudo-elements are not supported by css-select
  ::view-transition-new(header) -> Pseudo-elements are not supported by css-select
  ::view-transition-old(main) -> Pseudo-elements are not supported by css-select
  ::view-transition-new(main) -> Pseudo-elements are not supported by css-select
  ::view-transition-old(landing-image) -> Pseudo-elements are not supported by css-select
  ::view-transition-new(landing-image) -> Pseudo-elements are not supported by css-select
...
```

Char2sGu:
A very weird issue is that even if the Cloud Build is successful, the Rollout fails, after switching to `@apphosting/adapter-angular@17.2.11`.

- I use an Nx monorepo
- `@apphosting/adapter-angular@17.2.11` is installed at root package, as a prod dependency
- the `apphosting:build` script is at the root package

Lmk if you need any further context.

image
image

quedicesebas:
For Angular v19, we still need to set the npm task "`apphosting:build": "FRAMEWORK_VERSION=19.0.0 apphosting-adapter-angular-build"`?
Also, we still need to set:
```
"engines": {
    "node": "^20.11.1"
  }
```
???

PR #288, opened 2024-12-23
Allow firebase-admin 13.x

awong-dev:
Update peerDependencies so downstream dependencies have the option of using firebase-admin 13.x


leoortizz (changes requested):
Thanks @awong-dev! Before we can merge, please run `npm install` so that these changes are applied to the package-lock.json.

PR #304, opened 2025-02-24
Custom App Hosting Headers for NextJS

mathu97:
Update NextJS Adapter to:
1. Add `x-fah-adapter` header that indicates adapter name + version
   - Modifies the compiled `route-manifest.json` file in the standalone folder to add `x-fah-adapter` header.
3. Add `x-fah-middleware` header for apps that have a middleware.
   - Reades the compiled `middleware-manifest.json` file to check if there are any middlewares used by the app, and updates the `route-manifest.json` file to add `x-fah-middleware` header.

Manual Testing:
- [x] Next 15 with app router
- [ ] Next 14 with app router
- [x] Next 13 with app router
- [ ] Next 15 with pages router
- [ ] Next 14 with pages router
- [x] Next 13 with pages router

Depends on #305


packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -0,0 +1,41 @@
...
+      ...(middlewareExists(middlewareManifest)
+        ? [
+            {
+              key: "x-fah-middleware",
+              value: "true",
+            },
+          ]
+        : []),
+    ],
+    regex: "^(?:/((?:[^/]+?)(?:/(?:[^/]+?))*))?(?:/)?$",
```
annajowang, line 33: could there be a comment/example added of what this regex is trying to match?

packages/@apphosting/adapter-nextjs/src/interfaces.ts
```diff
@@ -102,10 +101,52 @@ export interface OutputBundleOptions {
...
   adapterVersion: string;
+}
+
+// Metadata schema for bundle.yaml outputted by next.js adapter
+export interface Metadata extends AdapterMetadata {
   framework: string;
   frameworkVersion: string;
 }
+
+// Metadata schema for asset bindings
```
Yuangwang, line 116: I don't think Metadata schema fits for all of these interfaces? I think these are all interfaces from the middleware-manifest.json file so it would be nice if that was clearer in the comments

packages/@apphosting/adapter-nextjs/src/interfaces.ts
```diff
@@ -102,10 +101,52 @@ export interface OutputBundleOptions {
...
+export interface AssetBinding {
+  filePath: string;
+  name: string;
+}
+
+// Metadata schema for middleware matchers
+export interface MiddlewareMatcher {
```
Yuangwang, line 123: Don't think this needs to be exported

packages/@apphosting/adapter-nextjs/src/interfaces.ts
```diff
@@ -102,10 +101,52 @@ export interface OutputBundleOptions {
...
+  regexp: string;
+  locale?: false;
+  has?: RouteHas[];
+  missing?: RouteHas[];
+  originalSource: string;
+}
+
+// Metadata schema for edge function definitions
+export interface EdgeFunctionDefinition {
```
Yuangwang, line 132: Also doesn't need to be exported as far as I can tell

packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -0,0 +1,41 @@
+import { AdapterMetadata, MiddlewareManifest } from "./interfaces.js";
+import { loadRouteManifest, writeRouteManifest, loadMiddlewareManifest } from "./utils.js";
+
+/**
+ * Adds the route overrides to the route manifest.
```
Yuangwang, line 5: Can you explain what this function achieves for App Hosting? Its not very clear what adding the route overrides to the route manifest does and why we need to do it.

packages/@apphosting/adapter-nextjs/src/bin/build.spec.ts
```diff
@@ -97,6 +98,10 @@ metadata:
       },
       path.join(tmpDir, ".next"),
       defaultNextVersion,
+      {
```
Yuangwang, line 101: Since this shows up a couple times, could you preinitialize it in the beforeeach section?
For example:

adapterMetadata = {  adapterPackageName: "@apphosting/adapter-nextjs",
        adapterVersion: "14.0.1",
      },
 
 Then 
 await generateBuildOutput(
      tmpDir,
      tmpDir,
      outputBundleOptions,
      path.join(tmpDir, ".next"),
      defaultNextVersion,
      adapterMetadata
 }

packages/@apphosting/adapter-nextjs/src/overrides.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    });
+
+    const updatedManifest = JSON.parse(
+      fs.readFileSync(routesManifestPath, "utf-8"),
+    ) as RoutesManifest;
+
+    assert.strictEqual(updatedManifest.headers.length, 1);
+
+    const headers = updatedManifest.headers[0];
+    assert.strictEqual(headers.source, "/:path*");
```
Yuangwang, line 115: Could you do this all in one go? Something like 
const expectedHeaders = {
  source: "/:path*",
  headers: [
    { key: "x-fah-adapter", value: "nextjs-1.0.0" },
    { key: "x-fah-middleware", value: "true" },
  ],
};

assert.deepStrictEqual(headers, expectedHeaders);

Same applies to previous test

packages/@apphosting/adapter-nextjs/e2e/app.spec.ts
```diff
@@ -114,4 +124,30 @@ describe("app", () => {
       "private, no-cache, no-store, max-age=0, must-revalidate",
     );
   });
+
+  it("should have x-fah-adapter header and no x-fah-middleware header on all routes", async () => {
```
jamesdaniels, line 128: can we manually codegen a middleware.ts in the test project that affects a couple of these routes?

packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -0,0 +1,41 @@
...
+  const middlewareManifest = loadMiddlewareManifest(appPath, distDir);
+  const routeManifest = loadRouteManifest(appPath, distDir);
+  routeManifest.headers.push({
+    source: "/:path*",
+    headers: [
+      {
+        key: "x-fah-adapter",
+        value: `nextjs-${adapterMetadata.adapterVersion}`,
+      },
+      ...(middlewareExists(middlewareManifest)
```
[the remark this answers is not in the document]
mathu97: I think that's what we eventually want to do, I believe we said for not we'll add this header if there is a middleware at all.

packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -0,0 +1,55 @@
...
+    ],
+    /* 
+      NextJs converts the source string to a regex using path-to-regexp (https://github.com/pillarjs/path-to-regexp) at 
+      build time: https://github.com/vercel/next.js/blob/canary/packages/next/src/build/index.ts#L1273.
+      This regex is then used to match the route against the request path.
+
+      This regex was generated by building a sample NextJs app with the source string `/:path*` and then inspecting the
+      routes-manifest.json file.
+    */
+    regex: "^(?:/((?:[^/]+?)(?:/(?:[^/]+?))*))?(?:/)?$",
```
jamesdaniels, line 47: btw you can use `path-to-regex` to generate this rather than hardcode

`> pathToRegexp("/:path*") => /^(?:\/((?:[^\/]+?)(?:\/(?:[^\/]+?))*))?(?:\/)?$/i`

jamesdaniels: oh see you have a comment explaining that now, heh

packages/@apphosting/adapter-nextjs/e2e/app.spec.ts
mathu97: Updated the `run-local.ts` file to support this. Would be great to have our e2e tests running against different next versions. I can look at that after this.

Yuangwang (approved):

jamesdaniels (approved):

PR #309, opened 2025-03-04
Next Config Override

mathu97:
Overrides the user's Next Config file to add configs optimized for Firebase App Hosting. Currently sets `images.unoptimized` to `true` unless user has it explicitly defined to `false` or has a custom image loader.


packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -1,5 +1,79 @@
...
+    } else if (fileExtension === "mjs") {
+      importStatement = `import originalConfig from './${originalConfigName}';`;
+    } else if (fileExtension === "ts") {
+      importStatement = `import originalConfig from './${originalConfigName.replace(".ts", "")}';`;
+    } else {
+      throw new Error(`Unsupported file extension: ${fileExtension}`);
+    }
+
+    // Create the new config content with our overrides
+    const newConfigContent = `${importStatement}
```
jamesdaniels, line 44: Let's add `// @ts-nocheck` so we don't fail the build if we violate any of their typescript rules

packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -1,5 +1,79 @@
...
+
+// This file was automatically generated by Firebase App Hosting adapter
+const config = typeof originalConfig === 'function' 
+  ? (...args) => {
+      const resolvedConfig = originalConfig(...args);
+      return {
```
jamesdaniels, line 50: let's make a helper for this return type so we don't duplicate logic

packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -1,5 +1,79 @@
...
+        ...resolvedConfig,
+        images: {
+          ...(resolvedConfig.images || {}),
+          unoptimized: true,
```
jamesdaniels, line 54: Don't override if they explicitly allow optimization, `unoptimized: false` and also we shouldn't turn off image optimization is they have manually specified a loader

packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -1,5 +1,79 @@
...
+
+  // Determine the file extension
+  const fileExtension = nextConfigFileName.split(".").pop() || "js";
+  const originalConfigName = `next.config.original.${fileExtension}`;
+  const newConfigName = `next.config.${fileExtension}`;
+
+  // Rename the original config file
+  try {
+    const originalContent = readFileSync(configPath, "utf-8");
+    await writeFile(join(projectRoot, originalConfigName), originalContent);
```
jamesdaniels, line 29: we can use `rename` instead of read/write

packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -1,5 +1,79 @@
...
+        unoptimized: true,
+      },
+    };
+
+${fileExtension === "mjs" ? "export default config;" : "module.exports = config;"}
+`;
+
+    // Write the new config file
+    await writeFile(join(projectRoot, newConfigName), newConfigContent);
+    console.log(`Successfully created ${newConfigName} with Firebase App Hosting overrides`);
```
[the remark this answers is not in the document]
mathu97: What do we want to do if it fails? Revert to old path? Throw a readable error?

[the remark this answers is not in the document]
jamesdaniels: I'd be inclined to fail the build with a sensible error but I'd be interested to hear what the rest of the team thinks.

packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -1,5 +1,98 @@
...
+  // Check if the file exists in the current working directory
+  const configPath = join(projectRoot, nextConfigFileName);
+
+  if (!(await exists(configPath))) {
+    console.log(`No Next.js config file found at ${configPath}`);
+    return;
+  }
+
+  // Determine the file extension
+  const fileExtension = nextConfigFileName.split(".").pop() || "js";
```
annajowang, line 26: what is the `js` for? 
are we expecting any format other than `next.config.[ts|js|mjs]`

packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -1,5 +1,98 @@
...
+    if (fileExtension === "js" || fileExtension === "cjs") {
+      importStatement = `const originalConfig = require('./${originalConfigName}');`;
+    } else if (fileExtension === "mjs") {
+      importStatement = `import originalConfig from './${originalConfigName}';`;
+    } else if (fileExtension === "ts") {
+      importStatement = `import originalConfig from './${originalConfigName.replace(".ts", "")}';`;
+    } else {
+      throw new Error(`Unsupported file extension: ${fileExtension}`);
+    }
+
```
annajowang, line 46: could a switch statement be used here instead? It should be easier to read

packages/@apphosting/adapter-nextjs/src/overrides.ts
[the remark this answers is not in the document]
annajowang: yeah i always prefer to fail early rather than try to be smart and fall back on something.
it makes the system easier to reason about, easier to debug, and less likely to confuse the user.

packages/@apphosting/adapter-nextjs/src/overrides.ts
jamesdaniels: there's also [extname](https://nodejs.org/api/path.html#pathextnamepath)

packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -1,5 +1,79 @@
...
+  if (!(await exists(configPath))) {
+    console.log(`No Next.js config file found at ${configPath}`);
+    return;
+  }
+
+  // Determine the file extension
+  const fileExtension = nextConfigFileName.split(".").pop() || "js";
+  const originalConfigName = `next.config.original.${fileExtension}`;
+  const newConfigName = `next.config.${fileExtension}`;
```
jamesdaniels, line 28: probably better to just reuse `nextConfigFileName`

packages/@apphosting/adapter-nextjs/src/overrides.ts
mathu97: Good catch, ya we only expect `ts|js|mjs`, we shouldn't default to 'js'.

annajowang (approved):
lgtm after github lint comment is resolved.

packages/@apphosting/adapter-nextjs/src/overrides.ts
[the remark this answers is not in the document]
mathu97: We now validate that override was done by checking existence of the config files (`next.config.[js|mjs|ts]` and `next.config.original.[js|mjs|ts]`), and that the resulting config is valid by calling Next's loadConfig again.

jamesdaniels (approved):

Issue #307, opened 2025-02-28
Build output can not be parsed everytime


alexdabast:
Hi,

We encounter some issues when trying to deploy our angular application.
It's a monorepo using NX and our package manager is PNPM instead of npm.

I found out the issue is realted to the parsing of the build output from the angular adapter

This line try to parse the build output from extractManifestOutput
https://github.com/FirebaseExtended/firebase-framework-tools/blob/aa0c4edbed4b34325f63cd797bf4be87168146c4/packages/@apphosting/adapter-angular/src/utils.ts#L111

This sometimes (not sure why exactly) fails as sometimes the output of the build can create (randomly ?) new lines

For example my last build created this output who is "valid" as the build didn't fail
```

> nx run web-shop:build:production
--
476 |  
477 | ❯ Building...
478 | ✔ Building...
479 | {
480 | "errors": [],
481 | "warnings": [],
482 | "outputPaths": {
483 | "root": "file:///
484 | workspace/dist/apps/ws",
485 | "browser": "file:///workspace/dist/apps/ws/browser",
486 | "server": "file:///workspace/dist/apps/ws/server"
487 | },
488 | "prerenderedRoutes": {}
489 | }
490 |  
491 |  
492 |  
493 | NX   Successfully ran target build for project web-shop
494 |  
495 |  
496 | undefined:10
497 | "root": "file:///
498 |  
499 | SyntaxError: Bad control character in string literal in JSON at position 2000
500 | at JSON.parse (<anonymous>)
501 | at parseOutputBundleOptions (file:///layers/google.nodejs.firebaseangular/npm_modules/node_modules/@apphosting/adapter-angular/dist/utils.js:87:33)
502 | at file:///layers/google.nodejs.firebaseangular/npm_modules/node_modules/@apphosting/adapter-angular/dist/bin/build.js:17:33
503 | at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
504 |  
505 | Node.js v20.18.3"}

````
As you can see the json is not valid because the output "terminal" create a new line (lines 483-484) to display the "root" output path ?!

I notice that this happen with pnpm but if I change to npm the issue do no appear.

One solution in the angular adapter would be to maybe remove every newline before parsing the result ?
The JSON would be valid even if the terminal generates multiples "new lines" to display the output

alexdabast:
Same bug happen with NPM 

[...]

PR #311, opened 2025-03-12
Make Angular manifest parsing more robust

Yuangwang:
This should fix the bug where people were occasionally the manifest file would add new line characters, breaking the parsing https://github.com/FirebaseExtended/firebase-framework-tools/issues/307

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25520,7 +25520,7 @@
       }
     },
     "packages/@apphosting/adapter-angular": {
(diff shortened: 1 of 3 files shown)


jamesdaniels (changes requested):
LGTM, I'd encourage some test coverage though. Would be nice if we can get the e2e test to trigger this behavior somehow.

jamesdaniels (approved):

PR #323, opened 2025-04-01
Set App Hosting overrides for NextJS apps without a Next Config

mathu97:


packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -104,30 +119,55 @@ function getCustomNextConfig(importStatement: string, fileExtension: string) {
...
-  const newConfigFileName = `next.config.original${originalConfigExtension}`;
-  const newConfigFilePath = join(root, newConfigFileName);
-  if (!(await exists(newConfigFilePath))) {
-    throw new Error(
-      `Next Config Override Failed: New Next.js config file not found at ${newConfigFilePath}`,
-    );
+  if (userNextConfigExists) {
+    // Ensure user's existing next config is preserved in a next.config.origin.* file
+    const originalConfigExtension = extname(originalConfigFileName);
+    const prservedConfigFileName = `next.config.original${originalConfigExtension}`;
```
annajowang, line 158: spelling "prserved"

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -30,16 +30,19 @@ const originalConfig = await loadConfig(root, opts.projectDirectory);
...
-  await validateNextConfigOverride(root, opts.projectDirectory, originalConfig.configFileName);
-}
+const userNextConfigExists = await exists(join(root, originalConfig.configFileName));
+await overrideNextConfig(root, originalConfig.configFileName, userNextConfigExists);
+await validateNextConfigOverride(
+  root,
+  opts.projectDirectory,
+  originalConfig.configFileName,
+  userNextConfigExists,
+);
```
annajowang, line 45: i'm not too sure about passing this extra userNextConfigExists, if we can avoid it. and since the value of `userNextConfigExists` determines which blocks of code get executed in each function, i think that means we can re-org the code a little so it's more predictable what happens in each function without having to read each line.

My initial question is do we even need to validate our own default / provided nextjs config? If not I think we could do something much more similar to the original code. Maybe:
```
const userNextConfigExists = await exists(join(root, originalConfig.configFileName));
if (userNextConfigExists) {
  await overrideNextConfig(root, originalConfig.configFileName);
  await validateNextConfigOverride(root, opts.projectDirectory, originalConfig.configFileName);
} else {
  // call some new function to write the DEFAULT_NEXT_CONFIG_FILE
}
```

Alternatively (my slight preference), we just do:

```
  await overrideNextConfig(root, originalConfig.configFileName);
  await validateNextConfigOverride(root, opts.projectDirectory, configFileName);
```
but `overrideNextConfig()` now handles two additional things:
(1) whether userNextConfigExists (and then writing DEFAULT_NEXT_CONFIG_FILE if not)
(2) ensuring that the preservedConfigFile still exists (instead of doing this in the validate function)

Then we don't have to pass around userNextConfigExists, validateNextConfig does not have to care if this is the original, if userconfigexists, etc.

also (2) relies on some details like knowing the original file prefix `next.config.original...`. So I think it would be good to keep any logic that requires knowing that prefix contained to one place.

you've thought about this more and I may be missing some challenges.

annajowang (approved):
lgtm with some comments

packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -10,26 +10,38 @@ import {
...
-  // Check if the file exists in the current working directory
-  const configPath = join(projectRoot, nextConfigFileName);
-
-  if (!(await exists(configPath))) {
-    console.log(`No Next.js config file found at ${configPath}`);
+  const userNextConfigExists = await exists(join(projectRoot, nextConfigFileName));
+  if (!userNextConfigExists) {
+    console.log(`No Next config file found, creating one with Firebase App Hosting overrides`);
+    try {
+      await writeFile(join(projectRoot, DEFAULT_NEXT_CONFIG_FILE), defaultNextConfigForFAH());
```
jamesdaniels, line 25: Does `nextConfigFileName` include the default filename? Then we can drop `nextConfigFileName` in favor.

jamesdaniels (approved):
Approve, nit in comment

PR #325, opened 2025-04-08
Only add x-fah-middleware header to routes that have middleware enabled

mathu97:
Currently `x-fah-middleware` header is being added to all routes of an app that has a middleware file.

This PR updates the logic to only add `x-fah-middleware` header to routes that have middleware enabled. This is done by  reading the `middleware-manifest.json` to get the routes that middleware is enabled on, and updating the `routes-manifest.json` file with the `x-fah-middleware` headers.

Testing this change locally:

1. Added a middleware.ts file that only matches the `/blah` route.
Screenshot 2025-05-02 at 12 13 54 PM

3. Build the app with the NextJs Adapter
4. Ran the app and saw the `x-fah-middleware` header only being added to the /blah route

Screenshot 2025-05-02 at 12 33 40 PM
Screenshot 2025-05-02 at 12 34 04 PM


packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -158,38 +161,47 @@ export async function addRouteOverrides(
...
+    routes-manifest.json file.
+  */
     regex: "^(?:/((?:[^/]+?)(?:/(?:[^/]+?))*))?(?:/)?$",
   });
 
-  await writeRouteManifest(appPath, distDir, routeManifest);
-}
+  // Add the middleware header to all routes for which middleware is enabled
+  const middlewareManifest = loadMiddlewareManifest(appPath, distDir);
+  const rootMiddleware = middlewareManifest.middleware["/"];
```
mathu97, line 188: @jamesdaniels My assumption here was that we only need to look at the root key as NextJs now only supports middleware at the root directory. It used to support nested middleware while in beta which is why I think the API is like this.

Similar logic in Opennext AWS adapter: https://github.com/opennextjs/opennextjs-aws/blob/73281c958d43d865fdded0d5d86dd82747365fee/packages/open-next/src/core/routing/util.ts#L160-L174

jamesdaniels: Yeah that's a good assumption

jamesdaniels (approved):

PR #346, opened 2025-06-26
Filter out node_modules

Yuangwang:
Filters out all directories other than .next/standalone, in practice this will pretty much only delete the extra node_modules in the root directory, improving build speeds

### DIFF
--- a/packages/@apphosting/adapter-nextjs/src/utils.ts
+++ b/packages/@apphosting/adapter-nextjs/src/utils.ts
@@ -194,9 +194,9 @@ async function generateBundleYaml(
     },
     outputFiles: {
       serverApp: {
-        include: [".next/standalone"]
-      }
-    }
+        include: [".next/standalone"],
+      },
+    },
   };
   await writeFile(opts.bundleYamlPath, yamlStringify(outputBundle));
   return;
--- a/packages/@apphosting/adapter-nextjs/src/utils.ts
+++ b/packages/@apphosting/adapter-nextjs/src/utils.ts
@@ -194,7 +194,7 @@ async function generateBundleYaml(
     },
     outputFiles: {
       serverApp: {
-        include: [".next/standalone"],
+        include: [".next/standalone"], // we're only including the standalone directory to exclude the node_modules folder specifically
       },
     },
   };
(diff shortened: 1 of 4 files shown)


packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -190,6 +192,11 @@ async function generateBundleYaml(
       framework: "nextjs",
       frameworkVersion: nextVersion,
     },
+    outputFiles: {
+      serverApp: {
+        include: [".next/standalone"],
```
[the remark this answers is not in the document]
Yuangwang: We’re actually moving everything else still, the only thing we’re removing with this change is the root dir’s node_modules.

Isn’t that extra in all cases so this shouldn’t break anyone?

sjjj986 (approved):

packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -151,7 +151,9 @@ async function moveResources(
   for (const path of pathsToMove) {
     const isbundleYamlDir = join(appDir, path) === dirname(bundleYamlPath);
     const existsInOutputBundle = await exists(join(outputBundleAppDir, path));
-    if (!isbundleYamlDir && !existsInOutputBundle) {
+    // Keep apphosting.yaml files in the root directory still, as later steps expect them to be there
```
annajowang, line 154: Assuming this is what caused the bug of outputFiles not being respected for nextjs.

packages/@apphosting/adapter-nextjs/src/utils.ts
annajowang, line 197: some comments within the code to explain context of why ".next/standalone" is hardcoded here would be good.

annajowang (approved):
one documentation nit

packages/@apphosting/adapter-nextjs/src/utils.ts
Yuangwang: Done

mehranabi:
It breaks monorepo setup.
We have our nextjs app at "apps/web" and our build output is at "apps/web/.next/standalone" and with this change our build goes through but the output is not copied to the final image.

PR #378, opened 2025-08-05
gitignore .apphosting/

annajowang:
part of making build adapters idempotent


packages/@apphosting/adapter-nextjs/src/utils.ts
```diff
@@ -203,6 +203,8 @@ async function generateBundleYaml(
   }
 
   await writeFile(opts.bundleYamlPath, yamlStringify(outputBundle));
+  const normalizedBundleDir = normalize(relative(cwd, opts.outputDirectoryBasePath));
```
Yuangwang, line 206: This function right now is in charge of generating just the bundle.yaml, can you move where the gitignore stuff happens to some other part of the code? Doing it in generateBuildOutput is fine I think

Yuangwang (approved):

PR #402, opened 2025-09-16
make local build callable

annajowang:
move main adapter build function to index.ts.

In the future we probably want to move some implementation details back into some internal file (not included in tsconfig.json).


annajowang:
> Just one question - how does this make local build callable?
it makes adapterbuild() directly importable / callable by the firebase CLI rather than making the firebase CLI call `npm exec apphosting-local-build`
index.ts is part of the "include" of tsconfig.js.
adapter-builds.ts is not. considered adding adapter-build.ts to the include list. but i think it's preferable to keep that list short and not include everything, later when more impelmentation logic / details is added, we should do them through helper functions not directly in the index.ts  file

PR #429, opened 2025-11-24
Add support for Node 24

### LINKED ISSUE #425: Node v24 support
As Node 24 is in LTS and firebase-tools was also just upgraded to support it, extending firebase-frameworks to support it as well woul be great.

service-paradis:
Closes #425 
Closes https://github.com/firebase/firebase-tools/issues/9564

Firebase supports Node 24. We should be able to deploy frameworks using this version.


rrussell986764-arch (approved):

packages/firebase-frameworks/package.json
```diff
@@ -43,7 +43,7 @@
     "dist"
   ],
   "engines": {
-    "node": "^16.0.0 || ^18.0.0 || ^20.0.0 || ^22.0.0"
+    "node": "^16.0.0 || ^18.0.0 || ^20.0.0 || ^22.0.0 || ^24.0.0"
```
leoortizz: I'd say we should [match firebase-tools](https://github.com/firebase/firebase-tools/blob/c65d86b27ead455837cdb7a6c7464d5cad579ead/package.json#L75-L77) and keep only `>=20.0.0 || >=22.0.0 || >=24.0.0`. @jamesdaniels WDYT?
```suggestion
    "node": ">=20.0.0 || >=22.0.0 || >=24.0.0"
```

leoortizz: Just confirming we need this change as the Firebase CLI itself errors out when trying to deploy with v16/18. @service-paradis FYI created a PR here https://github.com/firebase/firebase-tools/pull/9599 to handle it in the CLI too

```suggestion
    "node": "^20.0.0 || ^22.0.0 || ^24.0.0"
```

PR #451, opened 2025-12-04
Revert "Block builds on vulnerable nextjs versions"

annajowang:
Reverts FirebaseExtended/firebase-framework-tools#447 while updating the package.json


annajowang:
Not merging unless things go wrong

abhis3 (approved):

PR #670, opened 2026-07-28
feat(astro): support Astro 6 and 7

### LINKED ISSUE #667: Support Astro v6 in @apphosting/astro-adapter
## Summary

`@apphosting/astro-adapter@0.0.1` currently depends on `@astrojs/node@^9.x`, which is Astro 5 only. Astro 6 requires `@astrojs/node@^10.x` or `^11.x`. There is no `peerDependencies` constraint in the adapter, but the hard dependency on `@astrojs/node@^9` prevents Astro 6 from working correctly.

## Security motivation

Snyk flagged **SNYK-JS-ASTRO-17901336** ("Use of Non-Canonical URL Paths for Authorization Decisions", HIGH severity) against Astro v5. The fix is upgrading to Astro v6, but that is currently blocked by this adapter.

We assessed the risk for our specific app as low (auth is enforced at the page level via Firebase Admin SDK, not in Astro middleware), and have added a Snyk exception accordingly — but the upgrade is still needed.

## What needs to change

Looking at the compiled adapter source (`dist/index.js`), the change appears minimal:

1. **`package.json`** — bump `@astrojs/node` from `^9.0.1` to `^10.0.0 || ^11.0.0` and add `peerDependencies: { "astro": "^5 || ^6" }`
2. **`supportedAstroFeatures`** — verify `i18nDomains` and `envGetSecret` feature names/values are still valid in Astro v6's adapter API
3. **Publish a new version** — even a `0.0.2` would unblock the ecosystem

## Environment

- `@apphosting/astro-adapter`: `0.0.1`
- Current Astro: `5.18.1`
- Target Astro: `6.4.8` (latest)
- `@astrojs/node` latest: `11.0.2`
- Firebase App Hosting (App Hosting backend, `apphosting.yaml`)

## References

- Snyk advisory: https://snyk.io/vuln/SNYK-JS-ASTRO-17901336
- `@astrojs/node` changelog: https://github.com/withastro/astro/blob/main/packages/integrations/node/CHANGELOG.md

leoortizz:
## Add the App Hosting Astro adapter, with Astro 5, 6 and 7 support

Lands `@apphosting/astro-adapter` in `packages/@apphosting/experimental/` and extends
it to Astro 6 and 7. Fixes #667.

The adapter has lived on this branch since `0.0.1` was published to npm, and that
release only works on Astro 5. `@astrojs/node@10` (Astro 6) replaced the adapter
`args` config with a `virtual:astro-node:config` virtual module, so `0.0.1` fails to
build on Astro 6+.

### How version support works

The installed `@astrojs/node` major selects the config mechanism, since this adapter
reuses that package's server entrypoint:

| `@astrojs/node` | Astro | Mechanism |
| --- | --- | --- |
| 9 | 5 | adapter `args` export (unchanged from `0.0.1`) |
| 10, 11 | 6, 7 | `virtual:astro-node:config` module + `entrypointResolution: "auto"` |

`usesVirtualConfig()` resolves the installed major from the app's own `node_modules`,
so no user configuration changes. `UserOptions` is untouched — still just `mode`.

### Bug fixes

- **`.apphosting` was created with a non-recursive `mkdir`**, so any second build in
  the same directory failed with `EEXIST`. Now uses `ensureDir`.
- **`frameworkVersion` came from `npm view astro version`**, reporting npm's latest
  release rather than the installed one — every non-latest app was mislabelled, and it
  cost a network call per build. Now reads the installed `package.json`, matching how
  `firebase-tools` resolves Astro's version.

### Testing

Verified against real App Hosting deployments on all three majors — no mocks or
synthetic fixtures. Each was deployed and exercised over HTTP:

| | v5 (`@astrojs/node` 9) | v6 (10) | v7 (11) |
| --- | --- | --- | --- |
| SSR renders per request | ✅ | ✅ | ✅ |
| API routes | ✅ | ✅ | ✅ |
| Middleware executes | ✅ | ✅ | ✅ |
| `frameworkVersion` in bundle.yaml | `5.18.2` | `6.4.8` | `7.1.3` |

Each adapter option was toggled on real deployments to confirm it was load-bearing;
anything that worked without it was dropped. `createConfigPlugin` also supplies
`@astrojs/node`'s own defaults for `staticHeaders`, `bodySizeLimit` and
`experimentalDisableStreaming` — without them its server imports resolved to
`undefined`, silently removing the 1 GiB request body limit and emitting build
warnings in every user's project.

Also adds unit tests (`npm test`) covering the virtual config plugin.

### Notes

- Ships under `experimental/`; the `community/` workspace glob is dropped since this
  was its only member.
- `astro` and `@astrojs/node` are optional peer dependencies, matching the other
  adapters.


packages/@apphosting/experimental/adapter-astro/src/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+      handler() {
+        return RESOLVED_VIRTUAL_CONFIG_ID;
+      },
+    },
+    load: {
+      filter: { id: new RegExp(`^${RESOLVED_VIRTUAL_CONFIG_ID}$`) },
+      handler() {
+        return Object.entries(virtualConfig)
+          .map(([key, value]) => `export const ${key} = ${JSON.stringify(value)};`)
+          .join("\n");
```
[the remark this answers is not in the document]
leoortizz: Not true unless there's something I am missing, this would generate `export const client = undefined;` instead.

packages/@apphosting/experimental/adapter-astro/src/index.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  let _options: Options;
+  return {
+    name: "@apphosting/astro-adapter",
+    hooks: {
+      "astro:config:setup": ({ updateConfig, config }) => {
+        updateConfig({
+          image: {
+            endpoint: config.image.endpoint ?? "astro/assets/endpoint/node",
```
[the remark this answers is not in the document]
leoortizz: `AstroConfig["image"]` is non-optional on the resolved config

packages/@apphosting/experimental/adapter-astro/src/index.ts
```diff
@@ -0,0 +1,135 @@
...
+        setAdapter(getAdapter(_options));
+      },
+      "astro:build:done": async () => {
+        await ensureDir("./.apphosting");
+        const directoryName = dirname(fileURLToPath(import.meta.url));
+        const packageJsonPath = `${directoryName}/../package.json`;
+        if (!existsSync(packageJsonPath)) {
+          throw new Error(`Astro adapter package.json file does not exist at ${packageJsonPath}`);
+        }
+        const packageJson = JSON.parse(readFileSync(packageJsonPath, "utf-8"));
```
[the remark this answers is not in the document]
leoortizz: keeping as-is as it matches other adapters' use of sync APIs

.github/workflows/test.yml
Yuangwang, line 139: I think you need to define ADAPTER_ASTRO_NPM_TOKEN: ${{ secrets.ADAPTER_ASTRO_NPM_TOKEN }} here too and we need to set this up in the github as well

Yuangwang: I can help with setting it up on github

packages/@apphosting/experimental/adapter-astro/src/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Reads the version of a package as installed in the user's project.
+ * @param packageName The package to look up.
+ * @return The installed version, or undefined if its package.json cannot be read.
+ */
+export function getPackageVersion(packageName: string): string | undefined {
+  const require = createRequire(import.meta.url);
+
+  try {
+    const packageJsonPath = require.resolve(join(packageName, "package.json"), {
```
Yuangwang: I think this might cause some issues in windows with how the slashes are, probably want it to be const packageJsonPath = require.resolve(`${packageName}/package.json`) instead

.github/workflows/test.yml
leoortizz: just added it, let me know if I can help with anything else here

packages/@apphosting/experimental/adapter-astro/src/utils.ts
leoortizz: Good call thanks. I ended up blindly applying join in all paths

Yuangwang (approved):

PR #682, opened 2026-08-24
fix(nextjs): restore the user config when the override fails

### LINKED ISSUE #680: adapter-nextjs: next.config.mts is rejected, and the failure leaves the config renamed away
## Summary

`@apphosting/adapter-nextjs` rejects `next.config.mts`, a filename Next.js itself supports. `overrideNextConfig` switches on the config's file extension and throws for anything that is not `.js`, `.mjs`, or `.ts`. Since Next's own `loadConfig` discovers `.mts`, the adapter reaches that throw on any App Hosting build of a project that uses it.

Two aggravating details:

1. **The rename happens before the extension check**, and `overrideNextConfig` is called *outside* the `try…finally` that runs `restoreNextConfig`. The failure therefore leaves the project with its config renamed to `next.config.original.mts` and no `next.config.mts`, so no rollback.
2. Users have good reason to expect `.mts` to work: Next.js lists it in `CONFIG_FILES`, and Firebase CLI [15.9.1](https://firebase.google.com/support/release-notes/cli#version_1591_-_march_09_2026) announced "Added support for `next.config.ts` and `next.config.mts` in Next.js deployments". That support lives in `src/frameworks/next/` (the Hosting web frameworks path), so it does not reach App Hosting, but the release note reads as blanket support.

## Reproduction

Against the published package, no App Hosting deploy required:

```js
// npm i @apphosting/adapter-nextjs@14.0.21
// proj/next.config.mts contains a normal `export default nextConfig`
import { overrideNextConfig } from "@apphosting/adapter-nextjs/dist/overrides.js";
await overrideNextConfig("<abs>/proj", "next.config.mts");
```

```
before: next.config.mts

Overriding Next Config to add configs optmized for Firebase App Hosting
Error overriding Next.js config: Error: Unsupported file extension for Next Config: ".mts", please use ".js", ".mjs", or ".ts"

after:  next.config.original.mts      <- config renamed away, nothing written back
```

Verified on adapter 14.0.21 (current `latest`), Node 22, Next 16.3.1.

## Root cause

`packages/@apphosting/adapter-nextjs/src/overrides.ts` - the rename precedes the `switch`, whose `default` throws:

```ts
await renamePromise(configPath, originalPath);   // happens first
switch (fileExtension) {
  case ".js":  ...
  case ".mjs": ...
  case ".ts":  ...
  default: throw new Error(`Unsupported file extension for Next Config: "${fileExtension}", ...`);
}
```

`packages/@apphosting/adapter-nextjs/src/bin/build.ts` - the restore only guards the build, not the override:

```ts
if (await exists(nextConfigPath)) {
  await overrideNextConfig(root, nextConfig.configFileName);        // outside the try
  await validateNextConfigOverride(...);
}
try {
  await runBuild();
  ...
} finally {
  await restoreNextConfig(root, nextConfig.configFileName);
}
```

`nextConfig.configFileName` comes from Next's own `loadConfig`, and Next 16.3.1's `CONFIG_FILES` is `['next.config.js', 'next.config.mjs', 'next.config.ts', 'next.config.mts']`, so `.mts` is discovered and passed straight into the unsupported branch.

## Impact

- Any App Hosting Next.js app using `next.config.mts` fails its build.
- In Cloud Build the mutated workspace is ephemeral, so the rename is mostly invisible there. But the adapter is runnable locally (`npm exec apphosting-adapter-nextjs-build`), which is a reasonable way to reproduce adapter-only build failures, and there it mutates the developer's working tree and does not put it back.

## Suggested fix

1. Handle `.mts` alongside `.mjs`. `.mts` is ESM by extension regardless of the package's `type`, so the generated file needs the `export default config;` tail (emitting the `module.exports = config;` tail for it would produce a config with no default export. The import specifier needs a decision: the `.ts` case strips the extension, and whatever holds for `.ts` presumably holds for `.mts`.)
2. Independently of `.mts`: move the rename after the extension check, or bring `overrideNextConfig` inside the `try…finally`, so *any* failure after the rename restores the original config. Today a `writeFile` failure would strand the project the same way.

Om-singhaI:
`overrideNextConfig` renames the user's Next config before anything that can throw, and
`bin/build.ts` called it outside the `try` whose `finally` restores it. Any failure after the
rename left the project holding `next.config.original.<ext>` and no config at all, and a later
run does not recover it: `loadConfig` then reports the default `next.config.js`, so
`restoreNextConfig` looks for a backup under the wrong extension and returns early.

Cloud Build workspaces are ephemeral, but the adapter also runs locally through
`npm exec apphosting-adapter-nextjs-build`, where this mutates the developer's tree and does
not put it back.

The extension check and the config content now happen before the rename, the original is
restored if the write fails, and the override and its validation moved inside the `try`.

The new test covers the path that throws before the rename. The restore in the `catch` is
covered by inspection only, since making `writeFile` fail after a successful `rename` in the
same directory is not something I could drive from a unit test.

#684 shipped `.mts` support, which was the other half of #680.

Fixes #680

### DIFF
--- a/packages/@apphosting/adapter-nextjs/src/bin/build.ts
+++ b/packages/@apphosting/adapter-nextjs/src/bin/build.ts
@@ -35,6 +35,8 @@ const nextConfig = await loadConfig(root, opts.projectDirectory);
  *
  * We restore the user's Next Config at the end of the build, after the config file has been
  * copied over to the output directory, so that the user's original code is not modified.
+ * The override and its validation run inside the try block so that the restore in the
+ * finally block also covers a failure in either of them.
  *
  * If the app does not have a next.config.[js|mjs|ts|mts] file in the first place,
  * then can skip config override.
@@ -43,12 +45,13 @@ const nextConfig = await loadConfig(root, opts.projectDirectory);
  * one does not exist in the app's root: https://github.com/vercel/next.js/blob/23681508ca34b66a6ef55965c5eac57de20eb67f/packages/next/src/server/config.ts#L1115
  */
 const nextConfigPath = join(root, nextConfig.configFileName);
-if (await exists(nextConfigPath)) {
-  await overrideNextConfig(root, nextConfig.configFileName);
-  await validateNextConfigOverride(root, opts.projectDirectory, nextConfig.configFileName);
-}
 
 try {
+  if (await exists(nextConfigPath)) {
+    await overrideNextConfig(root, nextConfig.configFileName);
+    await validateNextConfigOverride(root, opts.projectDirectory, nextConfig.configFileName);
+  }
+
   await runBuild();
 
   const adapterMetadata = getAdapterMetadata();
--- a/packages/@apphosting/adapter-nextjs/src/overrides.spec.ts
+++ b/packages/@apphosting/adapter-nextjs/src/overrides.spec.ts
@@ -349,6 +349,26 @@ describe("next config overrides", () => {
     );
   });
 
+  it("should leave the original config in place when the override fails", async () => {
+    const { overrideNextConfig } = await importOverrides;
+    const originalConfig = `module.exports = { /* config options here */ }`;
+
+    // ".cjs" stands in for any extension the override does not know how to rewrite. If it
+    // ever becomes supported, pick another unsupported extension rather than deleting this.
+    fs.writeFileSync(path.join(tmpDir, "next.config.cjs"), originalConfig);
+
+    await assert.rejects(
+      async () => await overrideNextConfig(tmpDir, "next.config.cjs"),
+      /Unsupported file extension for Next Config/,
+    );
+
+    assert.equal(fs.readFileSync(path.join(tmpDir, "next.config.cjs"), "utf-8"), originalConfig);
+    assert.ok(
+      !fs.existsSync(path.join(tmpDir, "next.config.original.cjs")),
+      "No next.config.original.cjs backup should be left behind",
+    );
+  });
+
   it("should not do anything if no next.config.* file exists", async () => {
     const { overrideNextConfig } = await importOverrides;
     await overrideNextConfig(tmpDir, "next.config.js");
--- a/packages/@apphosting/adapter-nextjs/src/overrides.ts
+++ b/packages/@apphosting/adapter-nextjs/src/overrides.ts
@@ -28,41 +28,47 @@ export async function overrideNextConfig(projectRoot: string, nextConfigFileName
   const fileExtension = extname(nextConfigFileName);
   const originalConfigName = `next.config.original${fileExtension}`;
 
+  // Create a new config file with the appropriate import. This runs before the original
+  // config file is renamed, so an unsupported extension fails while the user's project is
+  // still untouched.
+  let importStatement;
+  switch (fileExtension) {
+    case ".js":
+      importStatement = `const originalConfig = require('./${originalConfigName}');`;
+      break;
+    case ".mjs":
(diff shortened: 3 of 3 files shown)


packages/@apphosting/adapter-nextjs/src/overrides.ts
```diff
@@ -28,41 +28,47 @@ export async function overrideNextConfig(projectRoot: string, nextConfigFileName
...
+      throw new Error(
+        `Unsupported file extension for Next Config: "${fileExtension}", please use ".js", ".mjs", ".ts", or ".mts"`,
+      );
+  }
+
+  // Create the new config content with our overrides
+  const newConfigContent = getCustomNextConfig(importStatement, fileExtension);
+
   // Rename the original config file
+  let originalConfigRenamed = false;
```
leoortizz: I think this can be a one-line change, which keeps the diff small and makes it easy to revert or pin down if something breaks after release. `restoreNextConfig` already returns early when there is no backup, so it's safe to call on every failure, and `originalConfigRenamed` isn't needed. The switch doesn't need to move above the rename either: if it throws after the rename, the restore in the catch undoes it. So this file can stay as it is on main, plus:

    } catch (error) {
      console.error(`Error overriding Next.js config: ${error}`);
      await restoreNextConfig(projectRoot, nextConfigFileName);
      throw error;
    }

Nice side effect: your `.cjs` test then actually runs the restore. Right now it throws before the rename, so the catch never gets exercised.

packages/@apphosting/adapter-nextjs/src/bin/build.ts
```diff
@@ -35,6 +35,8 @@ const nextConfig = await loadConfig(root, opts.projectDirectory);
  *
  * We restore the user's Next Config at the end of the build, after the config file has been
  * copied over to the output directory, so that the user's original code is not modified.
+ * The override and its validation run inside the try block so that the restore in the
+ * finally block also covers a failure in either of them.
```
leoortizz: nit: this comment restates the code, so I'd drop it

leoortizz (changes requested):
Thanks for the fix and the thorough write-up. I reproduced it locally, including a writeFile failure after the rename, and it works. One suggestion to make the change smaller before merging.

Om-singhaI:
Good call, that's a lot smaller. Pushed ca3d6f9.

`overrides.ts` is back to what's on main plus the one line in the catch, so `originalConfigRenamed` and the moved switch are both gone. You're right about the test as well: with the switch back below the rename, the `.cjs` case now goes through the catch, and taking the restore line out again makes that test fail on its own.

I kept the `build.ts` change. The override cleans up after itself now, but `validateNextConfigOverride` still runs outside the try, so a failure there ends the build with the user's config swapped out. Moving both inside lets the finally put it back. Dropped the comment.

31 unit tests pass in `@apphosting/adapter-nextjs`, and the diff is down to 26 lines added and 4 removed.

packages/@apphosting/adapter-nextjs/src/overrides.spec.ts
```diff
@@ -349,6 +349,26 @@ describe("next config overrides", () => {
     );
   });
 
+  it("should leave the original config in place when the override fails", async () => {
```
leoortizz: Nit: now that the switch is back below the rename, this case renames and restores rather than leaving the file untouched, so the title reads a bit off.

```suggestion
  it("should restore the original config when the override fails", async () => {
```

leoortizz (approved):

Yuangwang (approved):

